PR #2, opened 2016-02-25
Fixes unity script compile error 'TypeLoadException'

RodGreen:
Existing code:
`Type type = Type.GetType(CurrentResolverName, true);`
Can produce an error in the editor "TypeLoadException: Could not load type ResolverVer1_1."

Adjusting the code to use the AssemblyName "GooglePlayServices" fixes the issue.


RodGreen:
I signed it!
:D

claywilkinson:
Thanks!

PR #12, opened 2016-08-30
add firebase-iid to the should-explode list

alvinsay:
firebase-iid should be exploded


alvinsay:
I signed it!

stewartmiles:
Why add this to the list?  We already inspect the AARs to determine whether they should be exploded.

stewartmiles:
I just verified that firebase-iid automatically gets exploded when required, please reopen if you're still having an issue.

PR #13, opened 2016-08-30
Fixes code sample in SampleDependencies.cs

tkawachi:
Changed to assign to a static variable instead of a local variable.


stewartmiles:
Thanks

stewartmiles:
Oh it looks like you need to fix the conflicts in this change before we'll be able to merge it.

stewartmiles:
Turns out we already fixed this in 1e2ccb5fdb404d847f5782ddb48d82d2dd0eb46d .  I hope 1e2ccb5fdb404d847f5782ddb48d82d2dd0eb46d works for you.

Issue #16, opened 2016-10-10
Crashes Unity 5.5+


yuriy-universal-ivanov:
Hi unity-jar-resolver team,

Starting with Unity 5.5 (beta currently) there is no 11th BuildTarget (XBOX369), as result Unity crashes as soon as you import Play Services Resolver.

If you could fix it fast, it would be awesome.

Thanks in advance,
Yuriy.

stewartmiles:
This is resolved in the latest release.

PR #18, opened 2016-10-11
Fix: https://github.com/googlesamples/unity-jar-resolver/issues/16 - …

yuriy-universal-ivanov:
Fix for https://github.com/googlesamples/unity-jar-resolver/issues/16 - Crashes Unity 5.5+


yuriy-universal-ivanov:
I signed it!

stewartmiles:
Looking at this again it looks like we already resolved this issue with https://github.com/googlesamples/unity-jar-resolver/commit/d4f62659d85d488165222610a1c23acf27630f9a 

If you want to still apply your patch could you revisit it in the context of https://github.com/googlesamples/unity-jar-resolver/commit/d4f62659d85d488165222610a1c23acf27630f9a  to see whether it still makes sense?

PR #23, opened 2016-11-15
fix changelog markdown for github rendering

ekilah:
the h2 headers need to be spaced away from the unordered lists in order to render properly. also, indenting the unordered lists makes them appear like code samples instead of normal lists.


ekilah:
I signed it!

stewartmiles:
Thanks

stewartmiles:
It looks like you need to resolve the conflict before we can merge.

PR #24, opened 2016-11-16
Put dependencies in a distinguishable folder

rgngl:
This way it is easier to do clean-ups in no longer needed dependencies
and unity-jar-resolver copied dependencies stay separately from manually
placed Android native libraries.


rgngl:
I signed the CLA.

rgngl:
This breaks things with Unity 4. See PR #25 for a better solution.

PR #25, opened 2016-11-22
Add destination folder property for the resolver

rgngl:
Default is same as the previously hard-coded value.

### DIFF
--- a/source/PlayServicesResolver/src/PlayServicesResolver.cs
+++ b/source/PlayServicesResolver/src/PlayServicesResolver.cs
@@ -43,6 +43,11 @@ namespace GooglePlayServices
         private static IResolver _resolver;
 
         /// <summary>
+        /// Folder to copy the dependencies to
+        /// </summary>
+        private static string _destinationFolder = "Assets/Plugins/Android";
+
+        /// <summary>
         /// Flag used to prevent re-entrant auto-resolution.
         /// </summary>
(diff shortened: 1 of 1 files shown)


source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -43,6 +43,11 @@ public class PlayServicesResolver : AssetPostprocessor
         private static IResolver _resolver;
 
         /// <summary>
+        /// Folder to copy the dependencies to
+        /// </summary>
+        private static string _destinationFolder = "Assets/Plugins/Android";
```
stewartmiles: For this to work, the destination folder would need to at least be "Assets/Plugins/Android".  If you want this to be configurable why not use this as the prefix of the path and allow the user to configure the rest of the path e.g they could set the path to "Aars" which would place aars and exploded aars in "Assets/Plugins/Android/Aars"?

rgngl: I was just trying with Unity 5 now. `Assets/Plugins/Foo/Android` works, but `Assets/Plugins/Android/Foo` doesn't work. When the resolver puts the libraries in the latter in the inspector I can't even see the option checkboxes for platform and when building they're not included in Unity's staging area.

On the other hand, I haven't tried with Unity 4.x. Does this break there?

PR #30, opened 2016-12-07
Fix SDK path in PlayServicesSupport

rgngl:
EditorPrefs return empty string for missing keys, so the check in the
property misses this case. This causes build problems on CI environments
where there's no user setting for SDK path.

### DIFF
--- a/source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
+++ b/source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
@@ -97,7 +97,7 @@ namespace Google.JarResolver
         {
             get
             {
-                if (sdk == null)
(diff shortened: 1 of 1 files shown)


rgngl:
@claywilkinson Any comments about this?

rgngl:
@stewartmiles any comments about this?

source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
```diff
@@ -97,7 +97,7 @@ public string SDK
         {
             get
             {
-                if (sdk == null)
+                if (sdk == null || sdk.Length == 0)
```
rgngl: @stewartmiles yeah, that's better. done that.

stewartmiles:
Thanks for the commit.

PR #44, opened 2017-04-08
Fixes #43

### LINKED ISSUE #43: VersionHandler not working on machines that use comma separation instead of the dot
Basically, `float.TryParse()` will return false if the OS is in a language that uses comma as the decimal separator instead of the dot. This means that it will fall back to 5.4, which in turn causes it to attempt to load BuildTarget number 11, resulting in an instant crash.

bence98:
I have already signed the CLA.

### DIFF
--- a/source/VersionHandler/src/VersionHandler.cs
+++ b/source/VersionHandler/src/VersionHandler.cs
@@ -18,6 +18,7 @@ using UnityEngine;
 using UnityEditor;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System;
@@ -165,7 +166,7 @@ public class VersionHandler : AssetPostprocessor {
             if (!string.IsNullOrEmpty(version)) {
                 int dotIndex = version.IndexOf('.');
                 if (dotIndex > 0 && version.Length > dotIndex + 1) {
-                    if (!float.TryParse(version.Substring(0, dotIndex + 2), out result)) {
+                    if (!float.TryParse(version.Substring(0, dotIndex + 2), NumberStyles.Any, CultureInfo.InvariantCulture, out result)) {
                         result = 5.4f;
                     }
                 }


stewartmiles:
Thanks for the patch!

source/VersionHandler/src/VersionHandler.cs
```diff
@@ -165,7 +166,7 @@ private class FilenameComponents
             if (!string.IsNullOrEmpty(version)) {
                 int dotIndex = version.IndexOf('.');
                 if (dotIndex > 0 && version.Length > dotIndex + 1) {
-                    if (!float.TryParse(version.Substring(0, dotIndex + 2), out result)) {
+                    if (!float.TryParse(version.Substring(0, dotIndex + 2), NumberStyles.Any, CultureInfo.InvariantCulture, out result)) {
```
bence98: So, `float.TryParse()` is locale-sensitive by default. Thus, `float.TryParse("5.5"; out result)` will return true on an English machine, but false on a German one.

bence98: But Unity's version will always be in format "X.Y", which is not a valid float on German/Russian/etc. OSes. Therefore, `float.TryParse()` should ignore the locale.

bence98:
Will this be fixed then?
(Sorry if I sound too impatient, I'm just anxious about a project I have to release by the 28th, and this bug insta-crashes Unity 100% of the time, so I have to use another, poorly performing computer with an English OS)

bence98:
(Plus, I expect some delay between this bug getting fixed in JarResolver, and the AdMobile guys releasing a new plugin version, compiled against the fixed JarResolver)

stewartmiles:
Ok, gotcha.  I'll merge this now.   In terms of release delay I'm not sure what the schedule is @rampara do you have a sense of when you could merge this patch?

bence98:
Thank you! :)

rampara:
@bence98 I'm not sure I understand why you need a new version of the AdMob plugin. Doesn't building JarResolver with this merged fix and importing that into your Unity project resolve your issue? 

The [GMA Unity Gradle build script](https://github.com/googleads/googleads-mobile-unity/blob/master/build.gradle) does pull in the latest version of the Unity package from the root directory of this repo. Once that package is updated, you can build a new version of the GMA plugin that includes this fix.

bence98:
@rampara I am bad with Gradle, so I ended up importing the stock GMA 3.4 unitypackage, waited for Unity to crash&close, deleted PlayServicesResolver folder and imported PlayServicesResolver 1.2.16. It works like a charm, just was tedious. Thank you both for your time!

Issue #48, opened 2017-05-05
NullReferenceException with 1.2.19.0


yuriy-universal-ivanov:
Hi,

Due to https://github.com/googlesamples/unity-jar-resolver/issues/47 I had to update Play Services Resolver. Now it works, but all the time logs the exception in Unity Console:

NullReferenceException: Object reference not set to an instance of an object
GooglePlayServices.ResolverVer1_1.AarPathToPackageName (System.String aarPath)
GooglePlayServices.ResolverVer1_1.FindAarExplodeDataEntry (System.String aarPath)
GooglePlayServices.ResolverVer1_1.ShouldExplode (System.String aarPath)
GooglePlayServices.ResolverVer1_1.ShouldReplaceDependency (Google.JarResolver.Dependency oldDependency, Google.JarResolver.Dependency newDependency)
GooglePlayServices.PlayServicesResolver.<Resolve>m__C (Google.JarResolver.Dependency oldDependency, Google.JarResolver.Dependency newDependency)
Google.JarResolver.PlayServicesSupport.CopyDependencies (System.Collections.Generic.Dictionary`2 dependencies, System.String destDirectory, Google.JarResolver.OverwriteConfirmation confirmer)
GooglePlayServices.ResolverVer1_1.DoResolutionNoAndroidPackageChecks (Google.JarResolver.PlayServicesSupport svcSupport, System.String destinationDirectory, Google.JarResolver.OverwriteConfirmation handleOverwriteConfirmation)
GooglePlayServices.ResolverVer1_1+<DoResolution>c__AnonStorey8.<>m__10 ()
GooglePlayServices.ResolverVer1_1.DoResolution (Google.JarResolver.PlayServicesSupport svcSupport, System.String destinationDirectory, Google.JarResolver.OverwriteConfirmation handleOverwriteConfirmation, System.Action resolutionComplete)
GooglePlayServices.PlayServicesResolver.Resolve (System.Action resolutionComplete)
GooglePlayServices.PlayServicesResolver.AutoResolve ()
UnityEditor.EditorApplication.Internal_CallUpdateFunctions () (at /Users/builduser/buildslave/unity/build/artifacts/generated/common/editor/EditorApplicationBindings.gen.cs:249)

Resolver 1.2.19.0, Unity 5.6, macOS.

Thank you for any help with it,
Yuriy

claywilkinson:
[...]

PR #49, opened 2017-05-06
Fixing https://github.com/googlesamples/unity-jar-resolver/issues/48

yuriy-universal-ivanov:
Fix for https://github.com/googlesamples/unity-jar-resolver/issues/48

### DIFF
--- a/source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
+++ b/source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
@@ -502,12 +502,11 @@ namespace Google.JarResolver
 
                         // Extract the version from the filename.
                         // dep.Artifact is the name of the package (prefix)
-                        // The regular expression extracts the version number from the filename
-                        // handling filenames like foo-1.2.3-alpha.
-                        match = System.Text.RegularExpressions.Regex.Match(
-                            filenameWithoutExtension.Substring(
-                                dep.Artifact.Length + 1), "^([0-9.]+)");
-                        if (match.Success)
+                        string artifactVersion = ExtractVersionFromFileName(
+                            filenameWithoutExtension.Substring(dep.Artifact.Length + 1)
+                        );
+                        
+                        if (artifactVersion != null)
                         {
                             bool reportDependency = true;
                             // If the AAR is exploded and it should not be, delete it and do not
@@ -527,7 +526,6 @@ namespace Google.JarResolver
                             }
                             if (reportDependency)
                             {
-                                string artifactVersion = match.Groups[1].Value;
                                 Dependency currentDep = new Dependency(
                                     dep.Group, dep.Artifact, artifactVersion,
                                     packageIds: dep.PackageIds, repositories: dep.Repositories);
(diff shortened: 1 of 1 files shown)


yuriy-universal-ivanov:
I signed it!

image
Initially I used a different (non-google) primary email address with this GitHub account, but I've just switched the primary email account to the same as used when signing CLA.

source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
```diff
@@ -1038,7 +1036,10 @@ public void ClearDependencies()
                 KeyValuePair<Dependency, string> oldDepFilenamePair;
                 if (currentDepsByVersionlessKey.TryGetValue(dep.VersionlessKey,
                                                             out oldDepFilenamePair)) {
-                    if (oldDepFilenamePair.Key.BestVersion != dep.BestVersion &&
+                    string oldVersion = ExtractVersionFromFileName(
```
stewartmiles: Dependency.BestVersion should be the version string so in both cases this will be either a version string or an empty string see https://github.com/googlesamples/unity-jar-resolver/blob/master/source/JarResolverLib/src/Google.JarResolver/Dependency.cs#L158

Why did you need to change this to extract the version from the filename?  Are you seeing a file name sneak into this version field?

source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
```diff
@@ -502,12 +502,11 @@ public void ClearDependencies()
 
                         // Extract the version from the filename.
                         // dep.Artifact is the name of the package (prefix)
-                        // The regular expression extracts the version number from the filename
-                        // handling filenames like foo-1.2.3-alpha.
-                        match = System.Text.RegularExpressions.Regex.Match(
-                            filenameWithoutExtension.Substring(
-                                dep.Artifact.Length + 1), "^([0-9.]+)");
-                        if (match.Success)
+                        string artifactVersion = ExtractVersionFromFileName(
```
stewartmiles: Not sure why you need this.  The previous code attempted to extract the version from the filename and if the match succeeds then jumps into the block below and pulls the version from match.Groups[1].Value.  Did you see a problem in this code path or was this just a clean up?

source/JarResolverLib/src/Google.JarResolver/PlayServicesSupport.cs
```diff
@@ -1038,7 +1036,10 @@ public void ClearDependencies()
                 KeyValuePair<Dependency, string> oldDepFilenamePair;
                 if (currentDepsByVersionlessKey.TryGetValue(dep.VersionlessKey,
                                                             out oldDepFilenamePair)) {
-                    if (oldDepFilenamePair.Key.BestVersion != dep.BestVersion &&
+                    string oldVersion = ExtractVersionFromFileName(
+                        oldDepFilenamePair.Key.BestVersion);
+                    string newVersion = ExtractVersionFromFileName(dep.BestVersion);
+                    if ((oldVersion == null || (newVersion != null && oldVersion != newVersion)) &&
                         (confirmer == null || confirmer(oldDepFilenamePair.Key, dep))) {
```
stewartmiles: The stack trace reported in bug #48 starts here and bubbles up through to Resolver1_1.AarPathToPackageName where - from your report - the filename was null and therefore caused a null reference exception.  Which suggests https://github.com/googlesamples/unity-jar-resolver/blob/master/source/PlayServicesResolver/src/ResolverVer1_1.cs#L723 is returning a null path (as artifactPath) instead.

yuriy-universal-ivanov:
See https://github.com/googlesamples/unity-jar-resolver/issues/48#issuecomment-299999696

stewartmiles:
Ah ok, I see now.  Any chance you could update the commit message to reflect the comment you added here https://github.com/googlesamples/unity-jar-resolver/issues/48#issuecomment-299999696 ?

yuriy-universal-ivanov:
@stewartmiles ,
Sure. Done.

stewartmiles:
Thank you.

PR #57, opened 2017-05-26
get rid of warning when not on Android using runtime check

prime31:
Compile time check results in unreachable code warning on platforms other than Android


stewartmiles:
Thanks

PR #121, opened 2018-05-18
Added support for running ResolverVer1_1 synchronously

trondtactile:
The intention here is that projects who collect their own dependencies and requires the ability to run synchronously get the ability to do so by using the `DoResolution(PlayServicesSupport svcSupport, string destinationDirectory)` method in `ResoverVer1_1`, which is described as being the synchronous version of `DoResolutionUnsafe(svcSupport, destinationDirectory, unlockResolveAndSignalResolutionComplete)` in `DefaultResolver` (the base class).

### DIFF
--- a/source/PlayServicesResolver/src/ResolverVer1_1.cs
+++ b/source/PlayServicesResolver/src/ResolverVer1_1.cs
@@ -137,8 +137,6 @@ namespace GooglePlayServices
         private const int MajorVersion = 1;
         private const int MinorVersion = 1;
         private const int PointVersion = 0;
-
-        private bool runSynchronously;
         
         // Characters that are parsed by Gradle / Java in property values.
         // These characters need to be escaped to be correctly interpreted in a property value.
@@ -428,11 +426,12 @@ namespace GooglePlayServices
         /// <param name="androidSdkPath">Path to the Android SDK.  This is required as
         /// PlayServicesSupport.SDK can only be called from the main thread.</param>
         /// <param name="logErrorOnMissingArtifacts">Log errors when artifacts are missing.</param>
+        /// <param name="runSynchronously">Determines if the resolution should run synchronously.</param>
         /// <param name="resolutionComplete">Called when resolution is complete with a list of
         /// packages that were not found.</param>
         private void GradleResolution(
                 string destinationDirectory, string androidSdkPath,
-                bool logErrorOnMissingArtifacts,
+                bool logErrorOnMissingArtifacts, bool runSynchronously,
                 System.Action<List<Dependency>> resolutionComplete) {
             var gradleWrapper = Path.GetFullPath(Path.Combine(
                 gradleBuildDirectory,
@@ -845,6 +844,18 @@ namespace GooglePlayServices
         public override void DoResolution(
             PlayServicesSupport svcSupport, string destinationDirectory,
             System.Action resolutionComplete) {
+            DoResolutionInternal(svcSupport, destinationDirectory, false, resolutionComplete);
+        }
+
+        /// <summary>
+        /// Perform the resolution and the exploding/cleanup as needed synchronously.
+        /// </summary>
+        public override void DoResolution(PlayServicesSupport svcSupport, string destinationDirectory) {
+            DoResolutionInternal(svcSupport,destinationDirectory,true,() => {});
+        }
+
+        private void DoResolutionInternal(PlayServicesSupport svcSupport, string destinationDirectory, bool runSynchronously,
+            System.Action resolutionComplete) {
             // Run resolution on the main thread to serialize the operation as DoResolutionUnsafe
             // is not thread safe.
             System.Action resolve = () => {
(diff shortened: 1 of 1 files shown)


trondtactile:
Signed

source/PlayServicesResolver/src/ResolverVer1_1.cs
```diff
@@ -848,12 +858,26 @@ public override int Version()
...
+                {
+                    resolveUpdateQueue.Enqueue(resolve);
+                    EditorApplication.update += UpdateTryResolution;
+                }
             }
         }
 
+        public override void DoResolution(PlayServicesSupport svcSupport, string destinationDirectory)
+        {
+            runSynchronously = true;
```
stewartmiles: This seems a little dodgy to me as it's never set to false.  Once resolution has been run synchronously that's it all resolution will be synchronous.

source/PlayServicesResolver/src/ResolverVer1_1.cs
```diff
@@ -848,12 +858,26 @@ public override int Version()
...
+            if (runSynchronously)
+            {
+                resolve();
+            }else
+            {
+                lock (resolveLock)
+                {
+                    resolveUpdateQueue.Enqueue(resolve);
+                    EditorApplication.update += UpdateTryResolution;
+                }
```
stewartmiles: There are a lot of code paths that register with EditorApplication.update to do stuff in future on the main thread.  This makes me wonder whether it would be better to instead write a method that wraps PlayServicesResolver.updateQueue and allows everything to optionally run synchronously ?  I'm not sure why resolveUpdateQueue is separate to PlayServicesResolver.updateQueue to be honest.

source/PlayServicesResolver/src/ResolverVer1_1.cs
```diff
@@ -649,21 +651,29 @@ public override int Version()
...
+            else
+            {
+                // Run the build script to perform the resolution popping up a window in the editor.
+                var window = CommandLineDialog.CreateCommandLineDialog(
+                    "Resolving Android Dependencies");
+                window.summaryText = "Resolving Android Dependencies....";
+                window.modal = false;
+                window.progressTitle = window.summaryText;
+                window.autoScrollToBottom = true;
+                window.RunAsync(gradleWrapper, gradleArgumentsString,
```
stewartmiles: CommandLineDialog.RunAsync() is a superset of CommandLine.Run() would it make sense to have a mode where CommandLineDialog.RunAsync() wouldn't pop up a window to run a command instead of branching here?  There are loads of places CommandLineDialog is used in the process beyond this code path.

trondtactile: Not sure i understand. RunAsync starts a new thread, and invokes a callback when this thread is done. Run simple runs synchronously. 
Since i need to run synchronously, and RunAsync would be the wrong name if it in fact can run synchronously, then this doesn't seem to make much sense to me.

source/PlayServicesResolver/src/ResolverVer1_1.cs
trondtactile: Keep in mind that my patch was/is not intended to make the whole system support synchronous execution. I only focused on ResolverVer1_1, as that is the only part of this whole system we use. Making the whole system compatible is a task i feel is a bit too big for the changes we need.

source/PlayServicesResolver/src/ResolverVer1_1.cs
trondtactile: oh, i just assumed that one never reused an instance of the resolver. Will change it to pass the flag around as parameters instead then.

trondtactile:
Made some changes, have you had time to look through them?

source/PlayServicesResolver/src/ResolverVer1_1.cs
stewartmiles: Sorry about the slow response here.  I guess the point I was trying to make is that following the pattern that you're currently applying will end up duplicating all code paths that use CommandLineWindow.  To me this seems pretty unmaintainable.  If we want to support synchronous execution in this module then really everything should have the option of running synchronously.

Since it sounds like the problem you were originally trying to resolve was running resolution without the UI present (i.e Editor.update isn't called and windows don't work reliably) what I we can do instead is change CommandLineWindow to fall back to execution without a window and change the code which pumps the work queue (Editor.update) to instead have a general system for tasks run in the editor which can be configured to run everything synchronously if required.

trondtactile: Ok, i see your point. But i have to say, this sounds like a bigger task than we are willing to put time into for the time being. We will just use our fork for now ;)

trondtactile:
Aborted pull request due to time needed to fix the code up is just too much, when we can instead stay in this fork.

stewartmiles:
@trondtactile sorry about your need to fork.  We'll hopefully have a patch that fixes up all code paths soon.

PR #131, opened 2018-07-06
README improvements

### LINKED ISSUE #129: README Plugin Redistribution section issues
I think the "Plugin Redistribution" section of the README has a couple of issues.

1. The command given in the Plugin Redistribution section of the README is missing a path for the `-importPackage` argument. I'm guessing it should be something like `-importPackage ./play-services-resolver-1.2.46.0.unitypackage`?

2. There's a typo in this sentence: 

>The build of the Play Services Resolver plugin has Unity asset metadata that is configured **to so** that the all editor components...

3. The current text seems to be more about how to update an existing `Play Services Resolver` integration in a plugin project, without clobbering the `gvh` metadata. It doesn't really describe how to redistribute the plugin, nor in what situations I would want to do that. Is it as simple as including the `PlayServicesResolver` directory contents in my plugin? Is there ever an argument for *not* redistributing the plugin? As a Unity newbie it would help to have these details spelled out.

Thanks for reading.

s-hocking:
Fixes #129.

I'd like to improve the Plugin Redistribution Background section, but I'm not sure what it should say exactly. Currently it explains the technical details of why the `gvh_disable` option is required, but I don't think it's a very satisfactory explanation. 

- It doesn't explain what the impact on my project would be if I did not provide the option at import time.
- The Usage sections allude to `gvh_disable` being optional at import time, but the Plugin Redistribution section makes it sound mandatory.
- Is the option required at all if I don't want to use the Version Handler?

If the `gvh_disable` option is super-duper important in all situations (for both importing and exporting), then I will mention it in the various Usage sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,14 +36,17 @@ dependencies on Android) provides solutions for each of these problems.
 ## Android Dependency Management
 
 The *Android Resolver* component of this plugin will download and integrate
-Android libraries and handle any conflicts between plugins using shared
-libraries.  For example, if a Unity plugin `SomePlugin` requires the Google
-Play Games Android library and redistributes this library and its transitive
-dependencies in the folder `SomePlugin/Android/` when a user imports
-`SomeOtherPlugin` that includes the same libraries (potentially at a different
-version) in `SomeOtherPlugin/Android/` the developer using `SomePlugin` and
-`SomeOtherPlugin` will see a hard to interpret build error when building for
-Android.
+Android library dependencies and handle any conflicts between plugins that share 
+the same dependencies.
+
+Without the Android Resolver, typically Unity plugins bundle their AAR and 
+JAR dependencies, e.g. a Unity plugin `SomePlugin` that requires the Google
+Play Games Android library would redistribute the library and its transitive
+dependencies in the folder `SomePlugin/Android/`.  When a user imports
+`SomeOtherPlugin` that includes the same libraries (potentially at different
+versions) in `SomeOtherPlugin/Android/`, the developer using `SomePlugin` and
+`SomeOtherPlugin` will see an error when building for Android that can be hard 
+to interpret.
 
 Using the Android Resolver to manage Android library dependencies:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,18 +121,20 @@ target in the Unity editor.
 
 ## Usage
 
-   1. Add the unitypackage to your plugin project (assuming you are developing a
-      plugin).  Please read the [Plugin Redistribution](#plugin-redistribution)
-      section when you're considering how to export your package to share with
-      your users.
+   1. Add the `play-services-resolver-1.2.75.0.unitypackage` to your plugin
```
stewartmiles: Simply adding the plugin to your project isn't recommend as the Version Handler component will activate / enable all subcomponents of the plugin, preventing resolution of conflicting versions of this plugin.  This still needs a reference to the Plugin Redistribution section to ensure developers import the plugin correctly when redistributing the plugin.  I see Plugin Redistribution has been moved to bullet 3, would it make sense to add a note that step 3 is really important?

README.md
```diff
@@ -245,10 +251,8 @@ to the Xcode project exported by Unity.
 
 Dependencies for iOS are added by referring to CocoaPods.
 
-   1. Add the unitypackage to your plugin project (assuming you are developing a
-      plugin).  Please read the [Plugin Redistribution](#plugin-redistribution)
-      section when you're considering how to export your package to share with
-      your users.
+   1. Add the `play-services-resolver-1.2.75.0.unitypackage` to your plugin
```
stewartmiles: Similar to above, it would be great to make sure users don't miss the redistribution and also should probably avoid having specific version numbers in here.

s-hocking:
Thanks for the feedback. Since the Plugin Redistribution part is important, I've moved it closer to the start of the doc. I think it makes sense to be first, since importing the plugin is the first thing new developers will want to do. It's important they do it correctly, no?

README.md
```diff
@@ -110,26 +113,80 @@ Unity version 4.6.8 or higher.
...
+Redistributing the `Play Services Resolver` inside your own plugin will ease 
+the integration process for your users, by resolving dependency conflicts
+between your plugin and other plugins in a user's project.
+
+If you wish to redistribute the `Play Services Resolver` inside your plugin,
+you **must** follow these steps when importing the 
+`play-services-resolver-*.unitypackage`, and when exporting your own plugin
+package:
+
+   1. Import the `play-services-resolver-1.2.75.0.unitypackage` into your plugin
```
stewartmiles: Need to ditch the version number here as it's going to get out of date.

s-hocking: Whoops. Fixed now.

PR #162, opened 2018-10-31
Fixed incorrect dates in changelog.

Si1ver:
Some dates in changelog are incorrect.
Fixed this according to commit information. 
Note: the exact day are left as is, but year and month are corrected.


Si1ver:
I signed it!

stewartmiles:
Thanks

PR #173, opened 2018-12-14
Turning off auto resolution now prevents several methods from running in the background when they should now have been

oddgamesdaniel:
Methods that used to be hooked into the editor update tick are now able to be unhooked and will not automatically be assigned via the PlayServicesResolver's static constructor if auto resolution is turned off in the settings

### DIFF
--- a/source/PlayServicesResolver/src/PlayServicesResolver.cs
+++ b/source/PlayServicesResolver/src/PlayServicesResolver.cs
@@ -641,19 +641,34 @@ namespace GooglePlayServices
                 AndroidSdkRootChanged += ResolveOnAndroidSdkRootChange;
                 ScheduleAutoResolve();
             }
-
-            // Register events to monitor build system changes for the Android Resolver and other
-            // plugins.
-            RunOnMainThread.OnUpdate += PollBundleId;
-            RunOnMainThread.OnUpdate += PollBuildSystem;
-            RunOnMainThread.OnUpdate += PollAndroidAbis;
-            RunOnMainThread.OnUpdate += PollAndroidSdkRoot;
+			
+			if (GooglePlayServices.SettingsDialog.EnableAutoResolution)
+				LinkAutoResolution();
         }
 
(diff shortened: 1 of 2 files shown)


oddgamesdaniel:
CLA Signed

source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -641,19 +641,34 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
         }
 
-        /// <summary>
-        /// Called from PlayServicesSupport to log a message.
-        /// </summary>
-        internal static void LogDelegate(string message, PlayServicesSupport.LogLevel level) {
+		public static void UnlinkAutoResolution() {
+			// Unregister events to monitor build system changes for the Android Resolver and other
+			// plugins.
+			RunOnMainThread.OnUpdate -= PollBundleId;
```
stewartmiles: In general this all looks good.  However, Firebase and some other plugins rely upon the bundle ID notification to keep other configuration files up to date.  So turning off auto-resolution should probably not turn off bundle ID polling.  Any chance we could keep this unconditionally enabled?

oddgamesdaniel: From what I observed, the really heavy hitter was the polling for the android ABIs, so long as that doesn't run all the time the bulk majority of the allocations should go away. I'd trust your judgement on this one regarding polling the bundleid, I just removed any events that hooked into editor update tick. Not really sure how I can reformat my commit message, is that possible with Git?

stewartmiles: Yeah you can reformat your commit message with "git commit --amend", save a new commit message and force push the change to the head of your review repo.

stewartmiles:
@oddgamesdaniel I've done some reformatting of your commit and commit message but it's pretty much the same.  You're also still credited in the commit message.

Issue #202, opened 2019-04-12
Source pods with Resources


sdolim:
Source pods in project integration mode do not have their Resources directories scanned for resources.  In IOSResolver.UpdateProjectDeps(), only Resources directories inside .framework's are scanned.  (Our SDK is distributed in source form for business reasons.)

sdolim:
Interesting:  There is a script called Pods-Unity-iPhone-resources.sh in the Pods/Target Support Files/Pods-Unity-iPhone/ directory.  In a workspace-integrated project, the main target (Unity-iPhone) has a step in the Build Phases called "[CP] Copy Pods Resources" which invokes this script to copy all the resources into the build.  There is no corresponding Build Phase rule in the project-integrated Xcode project.  This must be why the Resources are not getting included.

sdolim:
Looks like this report is a dup of #71?

stewartmiles:
Fixed by Pull #204

Issue #201, opened 2019-04-12
Support for Development pods (cont)


sdolim:
Issue/PR #171 added support for development pods by allowing the "path" attribute in the dependencies file.  This is working great for workspace integration mode, but not for project integration mode.  The problem is that the function IOSResolver.UpdateProjectDeps() only scans the Pods/ subdirectory of the build, while development pods may point to paths outside of that file tree.  Therefore the development pods don't get scanned, so their files don't get added to the project.  (These files *are* listed in the Pods *project*, but that list is only used to filter the files found by scanning the Pods/ directory.)

We have customers whose build systems preclude them from using workspace integration mode, hence this is a nontrivial issue for our SDK.

stewartmiles:
@sdolim that does sound like a problem, any chance you're able to put a PR together for this?

sdolim:
Sure, I'll give it a go.

stewartmiles:
Fixed by pull #204 .

PR #204, opened 2019-04-17
Revise the way Xcode Project Integration works, in order to let cocoapods do more of the heavy lifting.

sdolim:
Basically, just add the Pods project to the Unity-iPhone project as a "subproject" (the way you used to do multiproject builds before workspaces), and add it to the build of the main target.  Also, don't suppress the pod command's normal project integration (i.e., don't add the line "install! 'cocoapods', :integrate_targets => false" to the Podfile) since that step does all of the work in setting up the build dependencies between the Unity-iPhone project and the Pods project.  That includes propagating build settings (search paths, flags, etc) from the pods into the main project.  (I believe this means the "bitcodeEnabled" attribute is now superfluous, but I didn't disturb the implementation of it for backwards compatibility.)

This change, aside from simplifying the code quite a bit, also lifts some restrictions that the old implementation had.  Local (i.e., "Development") pods that are located outside the Pods subdirectory are now supported.  Also, the resources in source pods are properly included in the app build.

Also leveraged some static calls provided by Unity's PBXProject class for finding the path project and main target name, so that logic is no longer needed in this code.

Addresses issues #201 and #202.


source/IOSResolver/src/IOSResolver.cs
```diff
@@ -1553,19 +1486,6 @@ public enum CocoapodsIntegrationMethod {
...
-        project.AddBuildProperty(target, "USER_HEADER_SEARCH_PATHS",
-                                 "$(PROJECT_DIR)/" + PODS_DIR + "/Headers/Public");
-        project.AddBuildProperty(target, "HEADER_SEARCH_PATHS",
-                                 "$(PROJECT_DIR)/" + PODS_DIR + "/Headers/Private");
-        project.AddBuildProperty(target, "FRAMEWORK_SEARCH_PATHS",
-                                 "$(inherited)");
-        project.AddBuildProperty(target, "FRAMEWORK_SEARCH_PATHS",
-                                 "$(PROJECT_DIR)/Frameworks");
-        project.AddBuildProperty(target, "LIBRARY_SEARCH_PATHS", "$(inherited)");
         project.AddBuildProperty(target, "OTHER_LDFLAGS", "-ObjC");
```
sdolim: Wasn't sure about these two.  If GTMSessionFetcher has a podspec / xcconfig that sets these, then the pods integration will take care of them, and these lines can also be removed.

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -1553,19 +1486,6 @@ public enum CocoapodsIntegrationMethod {
         project.ReadFromString(File.ReadAllText(pbxprojPath));
         string target = project.TargetGuidByName(TARGET_NAME);
         project.SetBuildProperty(target, "CLANG_ENABLE_MODULES", "YES");
```
sdolim: This ought to be taken care of by pods integration, but I wasn't sure if it was needed for GTMSessionFetcher (see comment below), so I left it in.

stewartmiles:
Ah well I should probably look at my Unity editor logs , I see this...

`MissingMethodException: Method not found: 'UnityEditor.iOS.Xcode.PBXProject.AddFile'.
Google.IOSResolver.OnPostProcessUpdateProjectDeps (BuildTarget buildTarget, System.String pathToBuiltProject) (at /Users/smiles/dev/src/unity-jar-resolver/source/IOSResolver/src/IOSResolver.cs:2127)
System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Reflection/MonoMethod.cs:222)
Rethrow as TargetInvocationException: Exception has been thrown by the target of an invocation.
System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Reflection/MonoMethod.cs:232)
System.Reflection.MethodBase.Invoke (System.Object obj, System.Object[] parameters) (at /Users/builduser/buildslave/mono/build/mcs/class/corlib/System.Reflection/MethodBase.cs:115)
UnityEditor.Build.BuildPipelineInterfaces+AttributeCallbackWrapper.OnPostprocessBuild (UnityEditor.Build.Reporting.BuildReport report) (at /Users/builduser/buildslave/unity/build/Editor/Mono/BuildPipeline/BuildPipelineInterfaces.cs:158)
UnityEditor.Build.BuildPipelineInterfaces+<OnBuildPostProcess>c__AnonStorey2.<>m__1 (IPostprocessBuildWithReport bpp) (at /Users/builduser/buildslave/unity/build/Editor/Mono/BuildPipeline/BuildPipelineInterfaces.cs:398)
UnityEditor.Build.BuildPipelineInterfaces.InvokeCallbackInterfacesPair[IPostprocessBuild,IPostprocessBuildWithReport] (System.Collections.Generic.List`1 oneInterfaces, System.Action`1 invocationOne, System.Collections.Generic.List`1 twoInterfaces, System.Action`1 invocationTwo, Boolean exitOnFailure) (at /Users/builduser/buildslave/unity/build/Editor/Mono/BuildPipeline/BuildPipelineInterfaces.cs:356)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr)
`

so it's failing while generating the Xcode project.  As I mentioned in the review, there are loads of versions of Xcode API floating around.  I recommend looking at the oldest most crusty version (e.g Unity 4 era) and seeing whether you can make things work with that.

stewartmiles:
Ok, I patched this again to use an overload of PBXProject.AddFile() that exists in more versions of Xcode API
https://github.com/stewartmiles/unity-jar-resolver/commit/dac11ab2d78cb4d42ac3f57e77192b99cacf7049

now Xcode projects are being transitively included in the core Xcode project .. very cool indeed, way less code.  I Like it!

This now seems to work for Firebase Auth, we'll need to go through and validate everything else.  Feel free to apply the patches I have in https://github.com/stewartmiles/unity-jar-resolver/tree/xcode-project-integration and squash them into your commit.  I think there is still some outstanding work in validating this in old Xcode API versions though.

sdolim:
Thanks for the feedback!  Most of the things you ran into were "code cleanups" that aren't really necessary to the fix (e.g., changing TARGET_NAME to a property), so I'm thinking of reverting those changes since the old code for those worked, it was just... messy.  I'll put in the AddFile() change as well.

stewartmiles:
@sdolim this looks ok, we'll try testing this out today. In the meantime could you squash the pull request into a single commit?

stewartmiles:
Thanks for squashing @sdolim we've finished testing with at least FIrebase - which provides pretty good coverage across different pod types  - so I'll merge this now.  I think the only risk here is whether old versions of Xcode don't support projects including projects, I vaguely remember there being a problem with xcproj files including xcproj files in the past which is why the approach we had in place was taken.

Again much appreciated!

Issue #217, opened 2019-05-17
prompt before auto resolution does not work as expected


Thaina:
I expect that `prompt before auto resolution` option would just launch popup to ask for skip when auto resolution would occur. But it just ask for enable or disable auto resolution and it always disable itself after popup once

Is it a bug? If it not I think it useless. Please add the actual functionality to let us skip auto resolution. It would be useful when we add many unity packages and then let it resolve after last package

stewartmiles:
@Thaina I believe this feature was added to enable folks to *disable* auto-resolution when importing the first of a series of packages, then they can enable auto-resolution or just force resolve when they're done.  You would prefer to see a prompt pop up for every auto-resolution attempt?  That would only then prompt the developer if the dependencies change again.

Thaina:
@stewartmiles Let me ask one thing

What is the time when we should go in the setting and toggle the `prompt before auto resolution` to on?

I thought it would be a time that we have nothing to import, but feel like there would be in the future. And so we must open the setting panel to toggle it for anytimes in the future

And if any dependency was changed by anything around this time, the popup will launch. We press anything and then it will disable itself

And then when there would be a time we need to import many packages. We try to import one, and if the last time we press enable, the resolution will kicking after we import the first package

If we need to remember to toggle it on after we import any package or has any package changed by any means. It would not be much difference to just disable it and just resolve it manually. So I think this feature is totally useless

It would be more useful if it always guard us off of the auto resolve that cause a friction and annoyance of importing multiple package then got stuck on resolving dependency

I think the user experience flow of the package importing is, if we have multiple package in the queue in our mind, we import one, and wait for import process finish to import another. We still having focus on our input device and so just another popup to ask that do we need to resolve dependency now or not is fine. We just click cancel and import next package, until the last package and we click yes

Thaina:
And also. If you would implement this feature. I would like to recommend that, this popup should have timer around 10 sec to just auto resolve

The use case of importing package is, we might just import large package and then AFK while enable this popup. So we would be more likely to have everything done when we get back

stewartmiles:
The use case is:
* Developer imports a plugin 1
* Resolution prompt pops up
* The developer is importing multiple plugins so disables auto-resolution
* Developer imports plugin 2..N
* After importing all plugins the developer goes to the Android Resolver settings, enables auto-resolution which then runs resolution.

[...]

PR #219, opened 2019-05-18
Delay auto resolve skippable dialog

Thaina:
For #217

### DIFF
--- a/source/PlayServicesResolver/src/PlayServicesResolver.cs
+++ b/source/PlayServicesResolver/src/PlayServicesResolver.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="PlayServicesResolver.cs" company="Google Inc.">
+// <copyright file="PlayServicesResolver.cs" company="Google Inc.">
 // Copyright (C) 2015 Google Inc. All Rights Reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
@@ -149,14 +149,12 @@ namespace GooglePlayServices {
                                 return true;
                             } else if (elementName == "package" &&
                                        parentElementName == "packages") {
-                                if (isStart && reader.Read() &&
-                                    reader.NodeType == XmlNodeType.Text) {
+                                if (reader.Read() && reader.NodeType == XmlNodeType.Text) {
                                     packages.Add(reader.ReadContentAsString());
                                 }
                                 return true;
                             } else if (elementName == "file" && parentElementName == "files") {
-                                if (isStart && reader.Read() &&
-                                    reader.NodeType == XmlNodeType.Text) {
+                                if (reader.Read() && reader.NodeType == XmlNodeType.Text) {
                                     files.Add(reader.ReadContentAsString());
                                 }
                                 return true;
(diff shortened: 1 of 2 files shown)


source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -928,32 +926,69 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
-                                // If autoResolveJobId is 0, ScheduleResolve()
-                                // has already been run and we should not run AutoResolve()
-                                // again.
-                                if (autoResolveJobId != 0) {
-                                    AutoResolve(() => {
-                                        lock (typeof(PlayServicesResolver)) {
-                                            autoResolving = false;
-                                            autoResolveJobId = 0;
+                if (autoResolving)
+					return;
```
stewartmiles: style: Could you move this to the previous line? 

We try to avoid:

if (foo)
  return;

in preference for

if (foo) {
  return;
}

or

if (foo) return;

The main issue with the continuation on a newline without braces is that other folks can come along and easily break the code with...

if (foo)
  do_something();
  return;

source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -928,32 +926,69 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
+
+				RunOnMainThread.Cancel(autoResolveJobId);
```
stewartmiles: style: indentation looks off here.  In general there are a lot of whitespace changes in this commit series.  Any chance you could:
- Squash the commit series into a single commit?
- Only apply whitespace changes in areas where you've modified how the code work?

That will make it much easier to see what's going on here.

source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -928,32 +926,69 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
+                autoResolveJobId = RunOnMainThread.Schedule(() => {
+                    lock (typeof(PlayServicesResolver)) {
+                        autoResolving = true;
+                    }
+
+                    DateTimeOffset delay = DateTimeOffset.Now.AddSeconds(Math.Min(30,GooglePlayServices.SettingsDialog.AutoResolutionDelay));
```
stewartmiles: style: We limit lines to 100 characters per line.  Any chance you could wrap this statement and other long lines?

source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -928,32 +926,69 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
+                    bool shouldResolve = true;
+                    AlertModal alert = null;
+                    RunOnMainThread.PollOnUpdateUntilComplete(() => {
+                        if(delay > DateTimeOffset.Now)
```
stewartmiles: style nit: "if (" rather than "if(".

source/PlayServicesResolver/src/PlayServicesResolver.cs
stewartmiles: nit: Would it make sense to call "delay", "resolveTime" since it's the time that resolution will occur?

source/PlayServicesResolver/src/PlayServicesResolver.cs
stewartmiles: In batch mode I'm pretty sure folks won't want a delay at all. 

ExecutionEnvironment.InBatchMode will tell you whether Unity is in non-interactive / batch mode.

source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -928,32 +926,69 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
+				RunOnMainThread.Cancel(autoResolveJobId);
+                autoResolveJobId = RunOnMainThread.Schedule(() => {
+                    lock (typeof(PlayServicesResolver)) {
+                        autoResolving = true;
+                    }
+
+                    DateTimeOffset delay = DateTimeOffset.Now.AddSeconds(Math.Min(30,GooglePlayServices.SettingsDialog.AutoResolutionDelay));
+                    bool shouldResolve = true;
+                    AlertModal alert = null;
+                    RunOnMainThread.PollOnUpdateUntilComplete(() => {
```
stewartmiles: What happens if a user uses the menu to start "Resolve" or "Force Resolve" when this job is running?

Typically the auto-resolve job is canceled here https://github.com/googlesamples/unity-jar-resolver/blob/master/source/PlayServicesResolver/src/PlayServicesResolver.cs#L1397 since the job Id isn't saved (it would also need to be in a critical section) this could result in multiple resolve jobs running at the same time.

source/PlayServicesResolver/src/SettingsDialog.cs
```diff
@@ -55,6 +56,7 @@ private class Settings {
                 verboseLogging = SettingsDialog.VerboseLogging;
                 autoResolutionDisabledWarning = SettingsDialog.AutoResolutionDisabledWarning;
                 promptBeforeAutoResolution = SettingsDialog.PromptBeforeAutoResolution;
+                autoResolutionDelay = Math.Min(Math.Max(0,SettingsDialog.AutoResolutionDelay),30);
```
stewartmiles: I'm curious why this would be limited to 30 seconds?

Thaina: I just think it should be limit to not let people shoot themselves in the foot by putting 9999+ delay time. 30 sec should be reasonable delay for popup if we don't stay away from keyboard

source/PlayServicesResolver/src/PlayServicesResolver.cs
Thaina: Doesn't that would be guarded at here? https://github.com/googlesamples/unity-jar-resolver/blob/e703601cec96ff60883b8999ca9aaa3ef2e459d9/source/PlayServicesResolver/src/PlayServicesResolver.cs#L944

I think if there would be any resolution interrupted, it would set `autoResolveJobId` to 0 at 
https://github.com/googlesamples/unity-jar-resolver/blob/e703601cec96ff60883b8999ca9aaa3ef2e459d9/source/PlayServicesResolver/src/PlayServicesResolver.cs#L1398 so if user use force resolve before the dialog was closed this job it would just be skip even user press OK

source/PlayServicesResolver/src/SettingsDialog.cs
stewartmiles: Ok, worth moving the "30" into a constant so that it's self documenting?

e.g MAXIMUM_AUTO_RESOLVE_DELAY_TIME

source/PlayServicesResolver/src/PlayServicesResolver.cs
stewartmiles: Looks like you still need to handle the batch mode case.

stewartmiles:
As I mentioned when I started looking at this, it would be nice to rebase and squash this change into a single commit to merge.

$ git rebase -i
# Squash all commits after the first one in the list so they're all combined into one
# Change the commit message to describe what this is doing (e.g https://chris.beams.io/posts/git-commit/ is a nice tutorial)
# Replace your branch with the squashed commit
$ git push origin -f HEAD:master

Thaina:
@stewartmiles Doesn't the git could squash whenever you merge my branch into your branch?

https://github.blog/2016-04-01-squash-your-commits/

Or I am misunderstand this?

Anyway I would fix everything as your comment first and would look about squashing when there was the last thing to be fixed

Thaina:
@stewartmiles I have tries to squash by tortoisegit, is this what you expected?

stewartmiles:
@Thaina typically I would use the command line to do this using `git rebase -i origin/master` that opens an editor and allows me to "squash" all commits in the series into a single commit i.e from origin/master..HEAD.

I *think* it's possible to do the same with tortoisegit https://stackoverflow.com/questions/12528854/how-to-perform-rebase-squash-using-tortoisegit

Thaina:
@stewartmiles I have already did that and the result is as I have committed here > d2974b3

stewartmiles:
I merged / rebased and squashed this commit into a local repo, the result is here https://github.com/stewartmiles/unity-jar-resolver/commit/fe13245ff9fb17118167baf3157dc000d2edd3a5

At the moment this doesn't really work.
- Since AlertModal uses EditorUtility.DisplayDialogComplex / EditorUtility.DisplayDialog it's not possible to update the dialog text for the count down.
- If resolution is skipped, auto-resolution doesn't appear to happen again
- After the timeout the alert dialog isn't closed

Any chance you built this and managed to test it?

Thaina:
I have fix it by using progressbar dialog. Now it could only cancel or wait to have it closed by itself

I could build the dll too. But it required to change something in gradle to let it be able to built in windows

PR #256, opened 2019-07-19
Fix for issue #79.

JimmyDeemo:
Check if the version control provider is active and enabled, and check out the file if so.

Issue #247, opened 2019-06-29
PlayServicesResolver 1.2.121 - iOS dll is broken


huran438:
iOS dll cant load into unity3d 2019.1.8f1

stewartmiles:
@huran438 what is broken exactly?  Could you share the error(s) you're seeing?

bdominguez:
Same problem here:

Unloading broken assembly Assets/PlayServicesResolver/Editor/Google.IOSResolver_v1.2.121.0.dll, this assembly can cause crashes in the runtime

stewartmiles:
That is spuriously reported by Unity but I'm fairly sure it will eventually load it anyway, are you sure it isn't loaded?  Can you see the menu `Assets > Play Services Resolver > iOS Resolver > Settings`?  If so, it's loaded and you can ignore the message from Unity.

https://issuetracker.unity3d.com/issues/unloading-broken-assembly-error-message-provides-insufficient-information

There are a load of bug reports about this sort of thing across a load of plugins and they're all caused by Unity throwing an error before compilation and plugin configuration is complete.

bdominguez:
@stewartmiles This error appears every time I compile not the first time you import the "*.unitypackage".

Personally when I have found that error is usually because of missing dependencies or incorrect target (.NET 3.5 vs .NET 4.x).

It's a real error, ignoring it is not an option because every time you recompile you get a new error in the console.

stewartmiles:
Any chance you have the Version Handler's version management disabled?

If not, perhaps your plugin import got into a bad state.  Try deleting Assets/PlayServicesResolver and reimporting the plugin.

bdominguez:
@stewartmiles I have created a fresh project and I have the same problem. Repro steps:

1. Create new project.
2. Import unity package.
3. Compilation finishes and I got an error on the console.

stewartmiles:
@bdominguez ok I've reproduce it, yeah it's when the iOS extension isn't installed in Unity the iOS DLL now fails to load in Unity 2019 :/ . This does not happen in earlier versions of Unity.

stewartmiles:
Workaround applied in https://github.com/googlesamples/unity-jar-resolver/releases/tag/v1.2.122

bdovaz:
@stewartmiles Thanks!!

bdominguez:
https://github.com/googlesamples/unity-jar-resolver/issues/247

PR #258, opened 2019-07-23
Version control checkout code around the AndroidResolverDependencies.xml file

JimmyDeemo:
Same as #256 but this time around the dependency state file.


source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -97,6 +97,15 @@ internal class DependencyState {
             public void WriteToFile() {
                 try {
                     Directory.CreateDirectory(Path.GetDirectoryName(DEPENDENCY_STATE_FILE));
+                    if (UnityEditor.VersionControl.Provider.enabled && UnityEditor.VersionControl.Provider.isActive &&
```
stewartmiles: Thanks again, since this code is the same boilerplate as the code in https://github.com/googlesamples/unity-jar-resolver/pull/256 could you add a static method to do this instead?

e.g

if (!CheckoutFile(DEPENDENCY_STATE_FILE)) {
  return;
}
...

Also, it would be useful to describe what went wrong if this fails.  At the moment it will silently fail which means that when we get a bug report that describes something not working we won't have any idea where it failed.

JimmyDeemo:
Sure, I’ll try and refactor it. Would you like it to throw or just log something?

stewartmiles:
Probably worth reporting an error and aborting without throwing unless you went and modified the callers to handle the exception.  Throwing an exception on the job queue can cause a chain of job failures.

JimmyDeemo:
I have refactored this code as suggested. Please let me know if you feel it need further refactoring.

JimmyDeemo:
Looks like this is a problem for the mainTemplate.gradle file too. As per my investigations into #247. I will endeavour to update this too.

stewartmiles:
@JimmyDeemo thanks for the update looks way better.

In terms of a commit to merge, any chance you could squash the commit series?  I noticed on the last pull request after I merged it that the series of patches was present, this is a pain in terms of rolling back when bisecting and doesn't really result in a self contained working patch in every commit.

Since you pointed out that mainTemplate.gradle is also an issue perhaps this is problematic for all file operations?  e.g the Android Resolver will delete libraries in the project when re-resolving, add new files, update existing files etc.  In addition, it can mutate maven repos in the project when patching them to support mainTemplate.gradle.

Any thoughts on how to handle large scale file additions, removals and changes?

JimmyDeemo:
I will take a look; i haven't had any problems when using the jar resolver for deletion of resolved libraries nor when library file have been added. It seems like patching existing items that are not check out is the problem. I'm going to try and solve this today as I need the patching to full work for #247.

JimmyDeemo:
> In terms of a commit to merge, any chance you could squash the commit series? 

Yes I can do this. Apologies, I haven't really worked with public projects before. Just to confirm, the idea behind this is that all of the Version Control stuff that would be merged from the pull request would be on a single commit?

JimmyDeemo:
I've added the checkout code to two more placed and tried to squash the commits into one. Please let me know if anthing else needed.

stewartmiles:
It looks like you still have a merge commit in the series which is a little odd.  I think you need to do the following...
- git checkout ${your_dev_branch}
- git pull origin/master # i.e pull from this repo
- git rebase -i # you can squash you commits into one in this step

then when making changes just do "git commit --amend" rather than "git commit" which will add changes to the commit you have out for review.  When pushing to the source repo (your branch) you can use "git push -f" to force an update of the commit in review.

Regarding the change contents.  I'm still a little concerned about code like this https://github.com/googlesamples/unity-jar-resolver/blob/56d476e23464f797d8c90b5fac0958af2e0887db/source/PlayServicesResolver/src/PlayServicesResolver.cs#L1394 and this method https://github.com/googlesamples/unity-jar-resolver/blob/56d476e23464f797d8c90b5fac0958af2e0887db/source/PlayServicesResolver/src/GradleResolver.cs#L1070 which hack around the Unity project without hitting source control.

JimmyDeemo:
I really messed up the history of this. I'm going to sort it out and resubmit, sorry.

Issue #262, opened 2019-07-25
Stopped polling job due to exception: System.MissingMethodException: UnityEditor.VersionControl.Task


Bibzball:
I updated Play Services Resolver to v1.2.123 and am getting this error on every auto-resolve.
Using Unity 2019.1.11f1

> `Stopped polling job due to exception: System.MissingMethodException: UnityEditor.VersionControl.Task UnityEditor.VersionControl.Provider.Checkout(string,UnityEditor.VersionControl.CheckoutMode)
>   at Google.ProjectSettings.Set[T] (System.String name, T value) [0x00024] in /Users/smiles/dev/src/unity-jar-resolver/source/VersionHandlerImpl/src/ProjectSettings.cs:119 
>   at Google.ProjectSettings+<SetBool>c__AnonStorey2.<>m__5 () [0x00000] in /Users/smiles/dev/src/unity-jar-resolver/source/VersionHandlerImpl/src/ProjectSettings.cs:154 
>   at Google.ProjectSettings.SavePreferences (Google.ProjectSettings+SettingsSave saveLevel, System.Action saveToProject, System.Action saveToEditor) [0x00035] in /Users/smiles/dev/src/unity-jar-resolver/source/VersionHandlerImpl/src/ProjectSettings.cs:135 
>   at Google.ProjectSettings.SetBool (System.String name, System.Boolean value) [0x00014] in /Users/smiles/dev/src/unity-jar-resolver/source/VersionHandlerImpl/src/ProjectSettings.cs:153 
>   at GooglePlayServices.SettingsDialog.set_UseJetifier (System.Boolean value) [0x00000] in /Users/smiles/dev/src/unity-jar-resolver/source/PlayServicesResolver/src/SettingsDialog.cs:225 
>   at GooglePlayServices.PlayServicesResolver.CanEnableJetifierOrPromptUser (System.String titlePrefix) [0x00179] in /Users/smiles/dev/src/unity-jar-resolver/source/PlayServicesResolver/src/PlayServicesResolver.cs:2270 
>   at GooglePlayServices.PlayServicesResolver.ResolveUnsafe (System.Action`1[T] resolutionComplete, System.Boolean forceResolution, System.Boolean isAutoResolveJob, System.Boolean closeWindowOnCompletion) [0x00012] in /Users/smiles/dev/src/unity-jar-resolver/source/PlayServicesResolver/src/PlayServicesResolver.cs:1610 
[...]

PR #263, opened 2019-07-26
Version control checkout code for various files.

JimmyDeemo:
This is a fixed version of #258 

Includes;
Static checkout function.
Checkout code for dependency file, project settings, mainTemplate.gradle and POM writing.

In the previous request you raised concerns about;

https://github.com/googlesamples/unity-jar-resolver/blob/56d476e23464f797d8c90b5fac0958af2e0887db/source/PlayServicesResolver/src/PlayServicesResolver.cs#L1394

and 

https://github.com/googlesamples/unity-jar-resolver/blob/56d476e23464f797d8c90b5fac0958af2e0887db/source/PlayServicesResolver/src/GradleResolver.cs#L1070

However, I have found that deleting and adding files to be fine (standard resolution works with .aar files) and so i'm loathed to add any checkout code here. The problem seems to be modifying existing files that could be write protected.

Hopefully this request is more suitable than the last. Please let me know of any problems.

### DIFF
--- a/source/PlayServicesResolver/src/GradleTemplateResolver.cs
+++ b/source/PlayServicesResolver/src/GradleTemplateResolver.cs
@@ -326,6 +326,12 @@ namespace GooglePlayServices {
                 }
                 outputLines.AddRange(currentOutputLines);
             }
+            if (!FileUtils.CheckoutFile(GradleTemplatePath))
+            {
+                PlayServicesResolver.Log(
+                    String.Format("Unable to checkou '{0}'.", GradleTemplatePath), level: LogLevel.Error);
+                return false;
+            }
             PlayServicesResolver.Log(
                 String.Format("Writing updated {0}", fileDescription),
                 level: LogLevel.Verbose);
--- a/source/PlayServicesResolver/src/PlayServicesResolver.cs
+++ b/source/PlayServicesResolver/src/PlayServicesResolver.cs
@@ -97,6 +97,13 @@ namespace GooglePlayServices {
             public void WriteToFile() {
                 try {
                     Directory.CreateDirectory(Path.GetDirectoryName(DEPENDENCY_STATE_FILE));
+                    if (!FileUtils.CheckoutFile(DEPENDENCY_STATE_FILE))
+                    {
+                        logger.Log(
+                            String.Format("Unable to checkout '{0}'. Dependency state has not been saved!",
+                                DEPENDENCY_STATE_FILE), LogLevel.Error);
+                        return;
+                    }
                     using (var writer = new XmlTextWriter(new StreamWriter(DEPENDENCY_STATE_FILE)) {
                             Formatting = Formatting.Indented,
                         }) {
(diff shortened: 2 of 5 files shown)


source/PlayServicesResolver/src/PlayServicesResolver.cs
```diff
@@ -97,6 +97,13 @@ internal class DependencyState {
             public void WriteToFile() {
                 try {
                     Directory.CreateDirectory(Path.GetDirectoryName(DEPENDENCY_STATE_FILE));
+                    if (!FileUtils.CheckoutFile(DEPENDENCY_STATE_FILE))
+                    {
+                        logger.Log(
+                            String.Format("Unable to checkout '{0}'. Dependency state has not been saved!",
+                                DEPENDENCY_STATE_FILE), LogLevel.Error);
```
stewartmiles: The problem here is that if this fails to save and auto-resolution is enabled, the resolver will get stuck in a loop attempting to resolve since the set of dependencies in memory won't match those in the state file.  I'll merge and clean fix this up.

source/VersionHandlerImpl/src/FileUtils.cs
```diff
@@ -288,5 +288,26 @@ internal class FileUtils {
...
+        /// </summary>
+        /// <param name="path">Path to the file that needs checking out.</param>
+        /// <returns>False should the checkout fail, otherwise true.</returns>
+        public static bool CheckoutFile(string path)
+        {
+            if (UnityEditor.VersionControl.Provider.enabled && UnityEditor.VersionControl.Provider.isActive &&
+                (!UnityEditor.VersionControl.Provider.requiresNetwork ||
+                 UnityEditor.VersionControl.Provider.onlineState == UnityEditor.VersionControl.OnlineState.Online))
+            {
+                var task = UnityEditor.VersionControl.Provider.Checkout(path,
```
stewartmiles: I'm not sure how this fixes https://github.com/googlesamples/unity-jar-resolver/issues/262 the code looks exactly the same which means it's possible we could end up with a MissingMethodException.  I'll merge and try working around this.

source/PlayServicesResolver/src/GradleTemplateResolver.cs
```diff
@@ -326,6 +326,12 @@ private class TextFileLineInjector {
                 }
                 outputLines.AddRange(currentOutputLines);
             }
+            if (!FileUtils.CheckoutFile(GradleTemplatePath))
+            {
+                PlayServicesResolver.Log(
+                    String.Format("Unable to checkou '{0}'.", GradleTemplatePath), level: LogLevel.Error);
```
stewartmiles: Typo, I'll clean up in the merge.

JimmyDeemo:
Sorry @stewartmiles, this wasn’t meant to address #262. I mainly wanted to clean up the previous pull request and to add more checkout calls.

I did not know how to solve the MissingMethodException at this time.

source/PlayServicesResolver/src/PlayServicesResolver.cs
JimmyDeemo: I see, I’ve actually seen that before and is one of the reasons we don’t enable auto resolution.

I think that if this file doesn’t save then it should probably fail resolution. I was unsure how to achieve this as I couldn’t return a failure from inside that function.

Issue #286, opened 2019-10-18
Support for Pod manipulation


pablo-wargaming:
Hi everyone,
I have a situation where I need to add a target to the Podfile generated (An App extension). I was able to modify the Podfile successfully in a post build script but since I can't execute it between the resolver's Podfile generation and pod install I have to manually do "pod install" after each build. I see 2 solutions to this problem that would help me:
1- Grant a way to add arbitrary text to the Podfile as part of the generation (I wasn't able to find any examples of how to do this)
2- Change [the values ](https://github.com/googlesamples/unity-jar-resolver/blob/master/source/IOSResolver/src/IOSResolver.cs#L395) of the execution order so that I can put something in between. The value must be integer and Pod generation is 4 while Pod install is 5. If you would multiply those by say 10 then people could add script executions in the desired order. 
Thanks!

stewartmiles:
Hi @pablo-wargaming 

It sounds reasonable to change the execution order values (i.e 2), as long as the execution order values as documented as part of the public API since changing these would be a breaking change.  Exposing an event property that is called from https://github.com/googlesamples/unity-jar-resolver/blob/2ad7cc90e5c48aa22fec7721678399e068e840f2/source/IOSResolver/src/IOSResolver.cs#L1693 may be an option but it's harder to use as the iOS Resolver isn't initially enabled when imported into a project so it could only be really called from a prebuilt DLL.

Happy to review / merge pull requests ;)

pablo-wargaming:
@stewartmiles here's the PR https://github.com/googlesamples/unity-jar-resolver/pull/288

stewartmiles:
Fixed in 1.2.130
https://github.com/googlesamples/unity-jar-resolver/releases/tag/v1.2.130

PR #288, opened 2019-10-21
Allow appending to generated podfile before install runs

pablo-wargaming:
https://github.com/googlesamples/unity-jar-resolver/issues/286
multiplied execution order attributes by 10 so podfile can be alterer after generation but before pod install runs


pablo-wargaming:
@googlebot I signed it!

Issue #284, opened 2019-10-12
Unity 2018.4.10f1 / Android Configuration Setup Error


JungPruna:
Hello.
I'm trying to use Firebase Authentication with GPGS, but cannot make them works together.

Tested with Unity 2018.4.10f1 new project.
Trying to "Google Play Games - Android Configuration" setup.

1.2.105 Fine (But error with recent firebase sdk when resolve)
1.2.123 Fine (But error with recent firebase sdk when resolve)
1.2.124 Fine (But error with recent firebase sdk when resolve) 
1.2.125 Invalid classname: Method VersionHandlerImpl.UpdateVersionedAssets not found
1.2.129 Invalid classname: Method VersionHandlerImpl.UpdateVersionedAssets not found

And sometimes "Ambiguous match found" appears. (Not sure how to reproduce)

ikitae:
me too

sonhoang2610:
This happened to me too

vnquanghvktqs:
me too, help me please

reddtoric:
During the installation/import of the plugins (GPGS, updated play service resolver, or firebase) that had the VersionHandlerImpl error, did the import window in Unity show any files (e.g. files in PlayServiceResolver) that are obsolete and will be replaced? If so, you have to cancel the import, delete the old files and then do your import. Somewhere in the docs in GPGS or the resolver, it says you have to delete old files before importing new ones.

vnquanghvktqs:
@reddtoric  : I used the latest PlayServiceResolver. Deleted all Google Play Games and re-enter it is still not possible. :(

vnquanghvktqs:
Resolved on play-services-resolver-1.2.124.0 release!.

firegate666:
I have the same issue in 1.2.129.0

andoprof:
Unity 2018.4.11f1
Firebase 6.6.0
Google Play Game Services 0.9.4, 0.10.4
Have same issue!
Please fix, thanks!

nishatoma:
Same issue here:
Unity 2019.2.7f2
Invalid classname: Method
VersionHandlerImpl.UpdateVersionedAssets not found

stewartmiles:
[...]

PR #289, opened 2019-10-23
Send forceUpdate parameter as positional arg instead of named arg to …

ozdemir08:
…prevent "VersionHandlerImpl.UpdateVersionedAssets not found".

[Play Games Plugin for Unity](https://github.com/playgameservices/play-games-plugin-for-unity) is depending on unity-jar-resolver. With the latest version v10.04 of PGS Plugin, developers started to see `Invalid classname: Method - VersionHandlerImpl.UpdateVersionedAssets` error, when they click Setup in Google Play Games - Android Configuration window. See the associated issues opened on [Play Games Plugin](https://github.com/playgameservices/play-games-plugin-for-unity/issues/2753) side and on [Unity Jar Resolver](https://github.com/googlesamples/unity-jar-resolver/issues/284) side.

I've tried different versions of Unity Jar Resolver with different versions of Play Games Plugin and the issue starts with Unity Jar Resolver v1.2.125. The reason for that is the change of signature of the method `UpdateVersionedAssets` [here](https://github.com/googlesamples/unity-jar-resolver/commit/6e214c8bacd36a247f3fb396a3295aac79451d94#diff-060ef1ed66ed6e2a143dd39e32026239L1841-R1871) in `VersionHandlerImpl.cs`, which causes ending up [here](https://github.com/googlesamples/unity-jar-resolver/blob/43d700803aa9e2680d9254ea706a3b6b6c2bc1da/source/VersionHandler/src/VersionHandler.cs#L540-L543), as `forceUpdate` is still given as `namedArgs` [here](https://github.com/googlesamples/unity-jar-resolver/blob/43d700803aa9e2680d9254ea706a3b6b6c2bc1da/source/VersionHandler/src/VersionHandler.cs#L382-L388). 

Sending forceUpdate parameter as a positional arg solves the issue. Another alternative could be changing the condition [here](https://github.com/googlesamples/unity-jar-resolver/blob/43d700803aa9e2680d9254ea706a3b6b6c2bc1da/source/VersionHandler/src/VersionHandler.cs#L525), but it will probably have implications which I haven't researched now.

### DIFF
--- a/source/VersionHandler/src/VersionHandler.cs
+++ b/source/VersionHandler/src/VersionHandler.cs
@@ -381,9 +381,7 @@ public class VersionHandler {
     /// </summary>
     public static void UpdateVersionedAssets(bool forceUpdate = false) {
         InvokeImplMethod("UpdateVersionedAssets",
-                         namedArgs: new Dictionary<string, object> {
-                             { "forceUpdate", forceUpdate }
-                         },
+                         args: new object[] { forceUpdate },
                          schedule: true);
     }
 


stewartmiles:
@ozdemir08 I'm glad this fixes the issue, I'm still puzzled as to why exactly this fixes the issue though since [this](https://github.com/googlesamples/unity-jar-resolver/blob/91e8010c475e784d7d1758f59961586229f0f2ec/source/VersionHandler/test/reflection/Assets/PlayServicesResolver/Editor/TestReflection.cs#L207) test case is supposed to validate this.

stewartmiles:
@ozdemir08 as discussed via video conference, I misread the method signature that was updated `VersionHandlerImpl.UpdateVersionedAssets` indeed has an incompatible overload when forceUpdate is specified.  Thanks for the patch!

PR #325, opened 2020-02-17
Unity OpenJDK path changed (2019.3.0, 2019.3.1)

unace:
add some lines to handle a new Unity OpenJDK path.
before
/Applications/Unity/Hub/Editor/{UnityVersion}/PlaybackEngines/AndroidPlayer/Tools/OpenJDK/{platform}
after
/Applications/Unity/Hub/Editor/{UnityVersion}/PlaybackEngines/AndroidPlayer/OpenJDK/

PR #329, opened 2020-02-29
Use the jdkRootPath provided by Unity

unace:
Since 2019.3 version, Unity has changed embedded JDK path and provided UnityEditor.Android.AndroidExternalToolsSettings.jdkRootPath API (https://docs.unity3d.com/2019.3/Documentation/ScriptReference/Android.AndroidExternalToolsSettings-jdkRootPath.html).
Previous MR(https://github.com/googlesamples/unity-jar-resolver/pull/325) wasn't enough to handle embedded JDK path, so I reverted it and changed codes to use Unity API.

### DIFF
--- a/source/PlayServicesResolver/src/JavaUtilities.cs
+++ b/source/PlayServicesResolver/src/JavaUtilities.cs
@@ -58,20 +58,19 @@ namespace GooglePlayServices {
                 // Unity 2019.x added installation of the JDK in the AndroidPlayer directory
                 // so fallback to searching for it there.
                 if (String.IsNullOrEmpty(javaHome) || EditorPrefs.GetBool("JdkUseEmbedded")) {
+#if UNITY_2019_3_OR_NEWER
+                    var openJdkDir = UnityEditor.Android.AndroidExternalToolsSettings.jdkRootPath;
+                    if (Directory.Exists(openJdkDir)) javaHome = openJdkDir;
+#else
                     var androidPlayerDir = PlayServicesResolver.AndroidPlaybackEngineDirectory;
                     if (!String.IsNullOrEmpty(androidPlayerDir)) {
                         var platformDir = UnityEngine.Application.platform.ToString().Replace(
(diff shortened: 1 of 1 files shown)


source/PlayServicesResolver/src/JavaUtilities.cs
```diff
@@ -58,20 +58,19 @@ internal class ToolNotFoundException : ApplicationException {
                 // Unity 2019.x added installation of the JDK in the AndroidPlayer directory
                 // so fallback to searching for it there.
                 if (String.IsNullOrEmpty(javaHome) || EditorPrefs.GetBool("JdkUseEmbedded")) {
+#if UNITY_2019_3_OR_NEWER
```
stewartmiles: We precompile the resolver into a DLL so any conditional logic needs to be conditional at runtime.

If it's _really_ only possible to change behavior based upon Unity version (it is super brittle), then you can use [ExecutionEnvironment.VersionMajorMinor](https://github.com/googlesamples/unity-jar-resolver/blob/ecfc41a15f4be10e8b3c88aca4352c4c17e38a3e/source/VersionHandlerImpl/src/ExecutionEnvironment.cs#L48)

Looking at the code below, I think you can check for the `UnityEditor.Android.AndroidExternalToolsSettings.jdkRootPath` property with reflection to determine the behavior rather than using the Unity version.

For examples, of how we're building accessors with reflection see https://github.com/googlesamples/unity-jar-resolver/blob/ecfc41a15f4be10e8b3c88aca4352c4c17e38a3e/source/PlayServicesResolver/src/UnityCompat.cs

unace: Thanks for the review, I will try reflection.

stewartmiles:
@unace FYI I have a patch in flight that takes this and uses reflection instead, I've also been updating the references to other tool paths to support Unity 2019.3+.

unace:
@stewartmiles sound great! thank you very much. I don't think this MR is not needed anymore. I will close it.

PR #335, opened 2020-03-10
Add ability to set XCode target from Dependencies file

### LINKED ISSUE #333: Resource bundles not linked when adding pod dependency to UnityFramework target
### Please fill in the following fields:
Unity editor version: 2019.3.3f
Play Services Resolver version: 1.2.136 (1.2.129+)
Features in Play Services Resolver in use (Android Resolver, iOS Resolver, VersionHandler, etc.): iOS Resolver
Plugins SDK in use (Firebase, Admob, Facebook, etc.): [AppLovin MAX Unity Plugin](https://github.com/AppLovin/AppLovin-MAX-Unity-Plugin)
Platform you are using the Unity editor on (Mac, Windows, or Linux): Mac
Platform you are targeting (iOS, Android, and/or desktop): iOS
Scripting Runtime (Mono, and/or IL2CPP): Mono

### Please describe the issue here:
Starting from 2019.3+ Unity has added a new target (`UnityFramework`) to the exported project and JAR Resolver (starting 1.2.129, Related Issue: https://github.com/googlesamples/unity-jar-resolver/issues/282) always adds the dependencies to this target. This has an unintended side effect of the resource bundles not being linked to the main target. Cocoapods is correctly adding these bundles to the `UnityFramework` but they are not being linked to the main target.

Adding the pod dependency directly to the `Unity-iPhone` target fixes this issue, but currently there is no way to specify the target to which the framework needs to be added to JAR Resolver.

### Please answer the following, if applicable:
What's the issue repro rate? (eg 100%, 1/5 etc) 100%
<br><br/>
It would be great if you could provide the ability to specify the desired target in the `Dependencies` file under the `iosPod` tag.

santoshbagadi:
Fixes #333

### DIFF
--- a/source/IOSResolver/src/IOSResolver.cs
+++ b/source/IOSResolver/src/IOSResolver.cs
@@ -52,6 +52,13 @@ public class IOSResolver : AssetPostprocessor {
         public string version = null;
 
         /// <summary>
+        /// The App's target to which to add the pod.
+        ///
+        /// See: XcodeTargetNames for valid target names.
+        /// </summary>
+        public string target = null;
+
+        /// <summary>
         /// Properties applied to the pod declaration.
         ///
         /// See:
@@ -158,6 +165,7 @@ public class IOSResolver : AssetPostprocessor {
         /// </summary>
         /// <param name="name">Name of the pod.</param>
         /// <param name="version">Version of the pod.</param>
+        /// <param name="target">The App's target to which to add the pod.</param>
         /// <param name="bitcodeEnabled">Whether this pod was compiled with
         /// bitcode.</param>
         /// <param name="minTargetSdk">Minimum target SDK revision required by
@@ -168,10 +176,12 @@ public class IOSResolver : AssetPostprocessor {
         /// a source.</param>
         /// <param name="propertiesByName">Dictionary of additional properties for the pod
         /// reference.</param>
-        public Pod(string name, string version, bool bitcodeEnabled, string minTargetSdk,
-                   IEnumerable<string> sources, Dictionary<string, string> propertiesByName) {
+        public Pod(string name, string version, string target, bool bitcodeEnabled,
+                   string minTargetSdk, IEnumerable<string> sources,
+                   Dictionary<string, string> propertiesByName) {
             this.name = name;
             this.version = version;
+            this.target = target;
             if (propertiesByName != null) {
                 this.propertiesByName = new Dictionary<string, string>(propertiesByName);
             }
@@ -209,6 +219,7 @@ public class IOSResolver : AssetPostprocessor {
             return pod != null &&
                    name == pod.name &&
                    version == pod.version &&
+                   target == pod.target &&
                    propertiesByName.Count == pod.propertiesByName.Count &&
                    propertiesByName.Keys.All(key =>
                        pod.propertiesByName.ContainsKey(key) &&
@@ -303,6 +314,7 @@ public class IOSResolver : AssetPostprocessor {
             var falseStrings = new HashSet<string> { "false", "0" };
             string podName = null;
             string versionSpec = null;
+            string target = null;
             bool bitcodeEnabled = true;
             string minTargetSdk = null;
             var propertiesByName = new Dictionary<string, string>();
@@ -327,6 +339,20 @@ public class IOSResolver : AssetPostprocessor {
                                 }
                             }
                             versionSpec = reader.GetAttribute("version");
+                            target = reader.GetAttribute("target");
+                            if (target == null) {
+                                target = DefaultXcodeTarget;
+                            } else {
+                                // Make sure that the provided target is supported.
+                                if (!XcodeTargetNames.Contains(target)) {
+                                    logger.Log(string.Format("Incorrect target name passed {0}." +
+                                                             "Adding the pod to the default" +
+                                                             "target: {1}",
+                                                            target,
+                                                            DefaultXcodeTarget));
+                                    target = DefaultXcodeTarget;
+                                }
+                            }
                             var bitcodeEnabledString =
                                 (reader.GetAttribute("bitcode") ?? "").ToLower();
                             bitcodeEnabled |= trueStrings.Contains(bitcodeEnabledString);
@@ -341,7 +367,9 @@ public class IOSResolver : AssetPostprocessor {
                                 return false;
                             }
                         } else {
-                            AddPodInternal(podName, preformattedVersion: versionSpec,
+                            AddPodInternal(podName,
+                                           target,
+                                           preformattedVersion: versionSpec,
                                            bitcodeEnabled: bitcodeEnabled,
                                            minTargetSdk: minTargetSdk,
                                            sources: sources,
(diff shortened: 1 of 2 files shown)


santoshbagadi:
@googlebot I signed it!

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -327,6 +338,16 @@ private class IOSXmlDependencies : XmlDependencies {
                                 }
                             }
                             versionSpec = reader.GetAttribute("version");
+                            target = reader.GetAttribute("target");
+                            if (target == null) {
+                                target = DefaultXcodeTarget;
+                            } else {
+                                // Make sure that the provided target is supported.
+                                if (!XcodeTargetNames.Contains(target)) {
+                                    logger.Log(string.Format("Incorrect target name passed {0}. Adding the pod to the default target: {1}", target, DefaultXcodeTarget));
```
stewartmiles: style: We keep all lines < 100 characters.

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -1808,6 +1843,7 @@ public enum CocoapodsIntegrationMethod {
                 }
                 AddPodInternal(
                     podName,
+                    "Unity-iPhone",
```
stewartmiles: Should this be the target name captured from line 1814?

i.e

"if (line.StartsWith("target 'Unity-iPhone' do") || line.StartsWith("target 'UnityFramework' do")) {"

I could imagine you could change this to a regex to extract the target name instead.
```
private static readonly Regex podFileTargetStartLine = new Regex("^target '([^']+)' do");
```

```
 while ((line = unityPodfile.ReadLine()) != null) {
  var currentTarget = "";
  var podFileTargetMatch = podFileTargetStartLine.Match(line);
  if (podFileTargetMatch.Success) {
    currentTarget = podFileTargetMatch.Groups[1].Value;
  }
  ...
```

You can then _potentially_ filter the target name with `XcodeTargetNames` and the target name to keep the pod in the correct target.

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -1799,26 +1799,44 @@ public enum CocoapodsIntegrationMethod {
...
 
         // We are only interested in capturing the dependencies "Pod depName, depVersion", inside
-        // of the specific target. However there can be nested targets such as for testing, so we're
+        // of the targets in XcodeTargetNames. However there can be nested targets such as for testing, so we're
         // counting the depth to determine when to capture the pods. Also we only ever enter the
         // first depth if we're in the exact right target.
         int capturingPodsDepth = 0;
+        string matchedTarget = null;
         var sources = new List<string>();
+        var targetsToCheck = new Dictionary<string, string>();;
```
stewartmiles: Ah, as mentioned on the prior match set I think a RegEx would simplify this quite a bit.

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -1843,7 +1862,7 @@ public enum CocoapodsIntegrationMethod {
                 }
                 AddPodInternal(
                     podName,
-                    "Unity-iPhone",
+                    matchedTarget ?? "Unity-iPhone",
```
stewartmiles: I'm a little confused by this, matchedTarget isn't present then something has gone wrong with the parsing.  Perhaps warn and don't add the pod to the wrong target instead?

source/IOSResolver/src/IOSResolver.cs
santoshbagadi: This definitely makes it a lot simple to get the target. Updated to use regex.

stewartmiles:
@santoshbagadi any chance you could squash this into a single commit so that it's easier to review in one go?

https://github.com/wprig/wprig/wiki/How-to-squash-commits

santoshbagadi:
@stewartmiles Done!

stewartmiles:
I just gave this a try with a FIrebase project and it appears to work, very nice job!

The only other thing I just noticed is that it would be great to update "documentation" (it's just a sample) in https://github.com/googlesamples/unity-jar-resolver/blob/master/sample/Assets/PlayServicesResolver/Editor/SampleDependencies.xml#L52

Also, feel free to tag your commit message with the issue this is resolving https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

santoshbagadi:
Added `target` to documentation. I'm not sure whether or not to mention what the default value will be, since it depends on the Unity version and could change in the future.

I've already tagged the issue that this PR resolves [here](https://github.com/googlesamples/unity-jar-resolver/pull/335#issue-385921719) 😅

stewartmiles:
@santoshbagadi sounds good, I just pushed a release so you'll need to rebase.  I think once that's done, I'll merge.

santoshbagadi:
Changes made

stewartmiles:
Sweet, thanks again!

stewartmiles:
@santoshbagadi while testing our latest release we found this caused a regression for iOS builds in some versions of Unity so we're rolling this back.

sp-ivan-hernandez:
This was fixing unity 2019.3 for us, not becaue we were able to set the target of a pod, but because in our xcode project the Unity-iPhone target in the Pods project existed again thanks to the
```
target Unity-iPhone do
end
```
added in the podfile (despite not having any pods).

stewartmiles:
@chkuang-g FYI

@sp-ivan-hernandez are you saying you need the empty "target" in the Podfile to fix an error for your build?  Any chance you have an example project that demonstrates this?

sp-ivan-hernandez:
i can not share the project right now but i can provide some info about it:

this is the Podfile

```
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '10.0'

target 'Unity-iPhone' do
end
target 'UnityFramework' do
  pod 'Bolts', '~> 1.7'
  pod 'FBSDKCoreKit', '~> 5.2'
  pod 'FBSDKLoginKit', '~> 5.2'
  pod 'FBSDKShareKit', '~> 5.2'
  pod 'Firebase/Analytics', '6.14.0'
  pod 'IronSourceAdColonyAdapter', '= 4.1.8.1'
  pod 'IronSourceAdMobAdapter', '= 4.3.8.4'
  pod 'IronSourceAmazonAdapter', '= 4.3.2.0'
  pod 'IronSourceAppLovinAdapter', '= 4.3.8.1'
  pod 'IronSourceFacebookAdapter', '= 4.3.10.1'
  pod 'IronSourceHyprMXAdapter', '= 4.1.3.4'
  pod 'IronSourceSDK', '= 6.14.0.0'
  pod 'IronSourceUnityAdsAdapter', '= 4.1.8.4'
  pod 'IronSourceVungleAdapter', '= 4.1.9.1'
  pod 'TapResearch', '= 2.0.9'
end
```

Without the "target 'Unity-iPhone' do end" the application was crashing on startup because of **IronSourceHyprMXAdapter** resources not found.

But with the "target 'Unity-iPhone' do end"  all is working fine.

Screenshot 2020-03-23 at 18 31 18

Notice that Pods-Unity-iPhone is created and Pods-Unity-iPhone-frameworks.sh adds the HyprMX.framework.

santoshbagadi:
@stewartmiles I have also noticed that having an empty `Unity-iPhone` target in the pod file was fixing the issue as well. 

Would you mind providing more information on what the exact issue that these changes are causing. Apart from adding an empty `Unity-iPhone` target to the pod file, it shouldn't have changed any other behavior when a target is not specified.

PR #355, opened 2020-04-13
Grand Cleanup of C# Project Files

Si1ver:
I've cleaned up C# project files.

What I changed and why:

- Added default hint paths to projects that references Unity or NUnit assemblies.
- All default hint paths are relative to solution file to make things easier. This fixes some paths in projects placed inside of another project had incorrect amount of `../`.
- Hint paths have trailing slashes to match default MSBuild style for directory variables. Gradle build file changed accordingly. This change might require one to update his local Gradle properties file.
- Assembly references to project assemblies are removed in favor of project references. This fixes problems when changed dependencies assemblies not copied to project output folder.
- `TargetFrameworkVersion` are set to `3.5` for all projects. This fixes build error in Rider,
- Added missing project references to `UnityPackageManagerResolverTests`
- Cleaned spacing and tabs.

Now if you have required Unity assemblies in `unity_dlls` folder, you will be able to build all projects in Visual Studio or Rider.

Hope that make contributing easier!


Si1ver:
I updated pull request to latest changes in master branch to solve merge conflicts and check everything again.

Williewill510 (changes requested):
What does this mean and how do I protect my data from being shared ???

Si1ver:
Hello!

This is outdated pull request that was not reviewed by developers for approx. three years.

If you have an questions about the changes in this PR, please be more specific, which part you are interested in. 

If you have questions about library itself, please ask them in "Issues" section.

Regards,
Anton

chrisyarbrough:
I apparently ran into the same issues as you did. When I opened the solution in Rider nothing compiled due to invalid references. I was able to fix them all in the same way you did and wonder why the project authors or most recent contributors didn't run into the same trouble. Is this different between Visual Studio and Rider or are others not using an IDE? I'd like to help clean things up, review or write documentation but it would be great if some of the more experienced maintainers could say something. :)

Si1ver:
Hi @chrisyarbrough !

I believe there are few reasons for this
1) I don't think that there are a lot of people who really need to recompile this project. Provided binaries are good enough for most of users 😃 
2) Rider is more strict in term of what's in .scproj files than Visual Studio. But it also not that common - most of developers are using Visual Studio. Maybe we are only two people who was trying to build this using Rider
3) The project build pipeline is based on old version of Unity. I don't remember exactly but I think I had to change something to make it compile using newer Unity version - the project just can't be fully built with default configuration with new Unity version.

Also I think that I was not careful enough and probably done too much changes in single PR, that's why this change was never accepted.

PR #394, opened 2020-08-07
physical paths for copying aars and normalized paths

vimanyu:
Normalizing file paths because we observed a mix of forward
and backward slashes on Windows. Also, using physical paths instead
of logical paths as that seems more stable for copy operations.


killerloader:
Sorry to bump, not sure if this is related, but I was also having issues where it couldn't copy some firebase files.
I tried the latest version of this but the issues still persists.
Eventually I figured out that it was because the file path lengths were too long, I simply moved the entire project to the root of my C drive and it fixed it. Hope this helps.

PR #399, opened 2020-08-31
Fixes #322, Use the raw property value when the enum does not contain this value

### LINKED ISSUE #322: Target API Level 29 Incompatibility
### Please fill in the following fields:
Unity editor version: 2019.3.0f6
Play Services Resolver version: 1.2.135
Features in Play Services Resolver in use (Android Resolver, iOS Resolver, VersionHandler, etc.):
Android Resolver
Plugins SDK in use (Firebase, Admob, Facebook, etc.): Admob
Platform you are using the Unity editor on (Mac, Windows, or Linux): Windows 10
Platform you are targeting (iOS, Android, and/or desktop): Android
Scripting Runtime (Mono, and/or IL2CPP): IL2CPP

### Please describe the issue here:
When I set my Target API Level to 29 (for Android 10 support), I get the following error message:

After setting the Target API to 29, pushing the "Resolve" or "Force Resolve" under the Android Resolver menu does nothing at all, whereas I'll get a "success" message on API Level 28. When changing the Target API to 28, there's no problem.

I'm using the NDK and SDK that come installed with Unity.

### Please answer the following, if applicable:
Repro rate 100%, happens every time I switch the Target API to 29 (which is what I want for Android 10 support).

Zyhlon:
This fixes #322 where the unity editor will lose its target sdk setting between unity restarts if you have >28 selected in 2019, since unity AndroidSdkVersions enum does not contain values above 28. 

I have no idea why the function for getting the enum value is so roundabout in the first place, perhaps someone could inform me of why we have to go through the enum type and the string name when we have the correct value from the beginning? Are we expecting the enum int values to diverge from the string values at some point? 

I made minimal impact on the function since I don't understand the purpose of it :)

### DIFF
--- a/source/AndroidResolver/src/UnityCompat.cs
+++ b/source/AndroidResolver/src/UnityCompat.cs
@@ -55,11 +55,22 @@ public class UnityCompat {
     }
 
     // Parses a UnityEditor.AndroidSDKVersion enum for a value.
-    private static int VersionFromAndroidSDKVersionsEnum(string enumName, string fallbackPrefKey,
+    private static int VersionFromAndroidSDKVersionsEnum(object enumValue, string fallbackPrefKey,
                                                          int fallbackValue) {
-        // If the enum property has no name it's not possible to parse the version so fallback to
-        // auto-selection.
-        if (String.IsNullOrEmpty(enumName)) return -1;
+        string enumName = null;
+        try {
+            enumName = Enum.GetName(typeof(AndroidSdkVersions), enumValue);
+        }
+        catch (ArgumentException)
+        {
+            //Fall back on auto if the enum value is not parsable
+            return -1;
+        }
+        
+        // If the enumName is empty then enumValue was not represented in the enum, 
+        // most likely because Unity has not yet added the new version,
+        // fall back on the raw enumValue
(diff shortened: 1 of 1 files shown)


Zyhlon:
@googlebot I signed it!

source/AndroidResolver/src/UnityCompat.cs
```diff
@@ -55,11 +55,22 @@ public class UnityCompat {
...
                                                          int fallbackValue) {
-        // If the enum property has no name it's not possible to parse the version so fallback to
-        // auto-selection.
-        if (String.IsNullOrEmpty(enumName)) return -1;
+        string enumName = null;
+        try {
+            enumName = Enum.GetName(typeof(AndroidSdkVersions), enumValue);
+        }
+        catch (ArgumentException)
+        {
```
chkuang-g, line 65:
```suggestion
        } catch (ArgumentException) {
```

chkuang-g (changes requested):
Looks good.  Thanks for the fix.

Just apply the suggested modification and it is good to go.

Zyhlon:
Done! Do you need me to squash or will you do it in the merge?

chkuang-g:
Thank you :)

Issue #384, opened 2020-06-16
Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7


peterpp:
### Please fill in the following fields:
Unity editor version: 2019.4.0f1 and also multiple of 2019.3.x versions
External Dependency Manager version: 1.2.155
Source you installed EDM: Unity Package Manager
Features in External Dependency Manager in use: Android Resolver
Plugins SDK in use: Firebase, Facebook, IronSource
Platform you are using the Unity editor on: Mac OS Catalina (10.15.5)
Platform you are targeting: Android
Scripting Runtime: IL2CPP

### Please describe the issue here:
I am unable to run Android Resolver because of exception:

```
[...]

PR #401, opened 2020-09-12
Update gradle version

MaxRev-Dev:
I had a repro of googlesamples/unity-jar-resolver#384 with 
 - Unity 2019.2.x 
 - External Dependency Manager 1.2.157
Updating gradle fixes the problem.


MaxRev-Dev:
@googlebot I signed it!

chkuang-g:
Hi @MaxRev-Dev 

Sorry I cannot accept this CL.  I tried this myself awhile ago and it broke EDM4U unit test.  Try this:
```
./gradlew testDownloadArtifacts
```

I commented my concerns here.
https://github.com/googlesamples/unity-jar-resolver/issues/384#issuecomment-695125140

Let me know if you can pass the test and make sure all SDK that is using EDM4U will not break because of the change.

PR #421, opened 2021-02-04
Fixing maven repo path separator char

marfx000:
The bug happens only if:
- "Patch mainTemplate.gradle" setting is enabled;
- You are resolving dependencies on Windows platform;

Examples of bad repo path string in mainTemplate.gradle:
file:///C:/Work/UnityProject\Assets/GeneratedLocalRepo/Firebase/m2repository

This happened because Path.Combine uses "\" as path separator for
windows platforms. But gradle actually expects not a regular path
but file URI, where "/" must be used as path separator.


marfx000:
@googlebot I signed it!

chkuang-g (approved):

Issue #239, opened 2019-06-20
Resolver repos are not source control friendly


bklooster:
Currently, when a plugin (such as Firebase's Unity plugin) defines its own repository for the library the resolver will create an entry in the resolver path that is user specific. This makes source control very painful because it is defined to the user's file path. 
![Screen Shot 2019-06-19 at 5 40 25 PM](https://user-images.githubusercontent.com/3290316/59810392-e0d26280-92b9-11e9-9316-e7806709d099.png)

Would it be possible to manually embed only the .aar files that are in their own repository so that the mainTemplate.gradle doesn't need to be hooked up to an absolute path name? The only work around 
Currently, to get around this problem, every user needs to force resolve on their machine prior to making a build. This is is very error prone. Also as part of the resolve process these files are recreated, which causes a change to appear in git due to the .meta files changing. 
![Screen Shot 2019-06-19 at 5 49 33 PM](https://user-images.githubusercontent.com/3290316/59810639-217eab80-92bb-11e9-94e8-cce03925a43c.png)

Any recommendations or fixes in new versions would be appreciated, thanks!

stewartmiles:
Looks like the problem is here:
https://github.com/googlesamples/unity-jar-resolver/blob/c1b77e391030a596697adaf57194160544fe0cc3/source/PlayServicesResolver/src/PlayServicesResolver.cs#L1885

Since the code injected calculates the project path it should probably be using the `unityProjectPath` variable if the repo is located in the project directory.

stewartmiles:
Oh now I'm confused, the resolver already does this here:
https://github.com/googlesamples/unity-jar-resolver/blob/c1b77e391030a596697adaf57194160544fe0cc3/source/PlayServicesResolver/src/PlayServicesResolver.cs#L1894

stewartmiles:
@bklooster which version are you using?

bklooster:
I just tested this with the latest I could find 1.2.118.0 and it is still happening. 

Thank you for the code link. 

Looking at the code you have a mechanic that says only do this relative path if export project is disabled. Our build process exports a Google project, and doesn't use the built-in gradle build system because company/project specific reasons (basically, we do some post-build modifications to the studio project similar to what we do with the xcode project) which needs to have the android studio project exported.

stewartmiles:
[...]

PR #438, opened 2021-05-13
Fixed invalid repo path patching & added user setting for path resolution

### LINKED ISSUE #426: [Bug] Invalid path to local repository when patching mainTemplate.gradle

  * Unity editor version: 2019.4.21f1
  * External Dependency Manager version: 1.2.164
  * Source you installed EDM4U: Unity Package Manager
  * Features in External Dependency Manager in use: Android Resolver, iOS Resolver
  * Plugins SDK in use: Firebase, AdMob, Facebook Audience Network, AppsFlyer, FairBid

_(Please list the full steps to reproduce the issue. Include device logs, Unity logs, and stack traces if available.)_

After updating Unity from 2019.4.9f1 to 2019.4.21f1, we needed to update our custom build pipeline:
We used to use [BuildOptions.AcceptExternalModificationsToPlayer](https://docs.unity3d.com/ScriptReference/BuildOptions.AcceptExternalModificationsToPlayer.html) in order to make Unity export a project. This was no longer a valid option, so we switched over to setting [EditorUserBuildSettings.exportAsGoogleAndroidProject](https://docs.unity3d.com/ScriptReference/EditorUserBuildSettings-exportAsGoogleAndroidProject.html) to `true`.

This has in turn caused the External Dependency Manager to patch the `mainTemplate.gradle` file with the absolute path to the generated local repository; this path uses an incorrect directory separator and will cause a gradle build failure.

See the backslash before "Assets":
```diff
        maven {
-            url (unityProjectPath + "/Assets/GeneratedLocalRepo/Firebase/m2repository")
+            url "file:///C:/Data/_Git/ExternalDependencyResolverIssue_InvalidPath\Assets/GeneratedLocalRepo/Firebase/m2repository"
        }
```

Building while the template is in this state produces the following gradle error:
```
FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Data\_Git\ExternalDependencyResolverIssue_InvalidPath\AndroidExport\unityLibrary\build.gradle' line: 11

* What went wrong:
Could not compile build file 'C:\Data\_Git\ExternalDependencyResolverIssue_InvalidPath\AndroidExport\unityLibrary\build.gradle'.
> startup failed:
  build file 'C:\Data\_Git\ExternalDependencyResolverIssue_InvalidPath\AndroidExport\unityLibrary\build.gradle': 11: unexpected char: '\' @ line 11, column 82.
     dencyResolverIssue_InvalidPath\Assets/Ge
                                   ^

  1 error
```

What's the issue repro rate? (eg 100%, 1/5 etc)
100%

**What happened? How can we make the problem occur?
This could be a description, log/console output, etc.**

Make sure the project contains a `mainTemplate.gradle` file, and that External Dependency Manager is set up to patch this file.
I've attached a project with a highly simplified setup:
[ExternalDependencyResolverIssue_InvalidPath.zip](https://github.com/googlesamples/unity-jar-resolver/files/6109532/ExternalDependencyResolverIssue_InvalidPath.zip) (Unity 2019.4.21f1)

1. In File > Build Settings > Android, enable "Export Project"
2. Save the project.
3. Observe that the path in `Plugins\Android\mainTemplate.gradle` is incorrectly patched.

I've exported an Android build in the `AndroidExport` folder in the zip.
Run `gradlew build` and you will see it fail as described above.

If the "Export Project" option is disabled, `mainTemplate.gradle` is patched back to a valid state.

Note: I would highly prefer that the fix for this issue is to not change the value to an absolute path, as it is a pain to work with in source control. This has also been mentioned in issue #239

strawlink:
This fixes #426 where paths would be combined incorrectly on Windows (used backslash when forward slash is expected).

As discussed in the comments, I also added a user setting "Use relative repository path":
image

The value defaults to `false` to preserve existing behavior.

### DIFF
--- a/source/AndroidResolver/src/PlayServicesResolver.cs
+++ b/source/AndroidResolver/src/PlayServicesResolver.cs
@@ -522,6 +522,16 @@ namespace GooglePlayServices {
             }
         }
 
+        /// <summary>
+        /// Whether the Gradle template should use the relative or absolute path to the local repository.
+        /// Only relevant when <see cref="GradleProjectExportEnabled"/> is also enabled.
+        /// </summary>
+        public static bool UseRelativeRepoPath {
+            get {
+               return SettingsDialogObj.UseRelativeRepoPath;
+            }
+        }
+
         // Backing store for GradleVersion property.
         private static string gradleVersion = null;
         // Extracts a version number from a gradle distribution jar file.
@@ -2106,6 +2116,7 @@ namespace GooglePlayServices {
             var lines = new List<string>();
             if (dependencies.Count > 0) {
                 var exportEnabled = GradleProjectExportEnabled;
+                var useRelativeRepoPath = UseRelativeRepoPath;
                 var projectPath = FileUtils.PosixPathSeparators(Path.GetFullPath("."));
                 var projectFileUri = GradleResolver.RepoPathToUri(projectPath);
                 lines.Add("([rootProject] + (rootProject.subprojects as List)).each { project ->");
@@ -2140,9 +2151,9 @@ namespace GooglePlayServices {
                             repoPath = relativePath;
                         }
 
-                        // If "Export Gradle Project" setting is enabled, gradle project expects
-                        // absolute path.
-                        if (exportEnabled) {
+                        // If "Export Gradle Project" setting is enabled, use absolute path
+                        // unless user has actively opted to use relative path
+                        if (exportEnabled && !useRelativeRepoPath) {
                             repoUri = String.Format("\"{0}\"",
                                                     Path.Combine(projectFileUri, repoPath));
                         } else {
@@ -2486,6 +2497,7 @@ namespace GooglePlayServices {
                 {"explodeAars", SettingsDialogObj.ExplodeAars.ToString()},
                 {"patchAndroidManifest", SettingsDialogObj.PatchAndroidManifest.ToString()},
                 {"patchMainTemplateGradle", SettingsDialogObj.PatchMainTemplateGradle.ToString()},
+                {"useRelativeRepoPath", SettingsDialogObj.UseRelativeRepoPath.ToString()},
                 {"localMavenRepoDir", SettingsDialogObj.LocalMavenRepoDir.ToString()},
                 {"useJetifier", SettingsDialogObj.UseJetifier.ToString()},
                 {"bundleId", GetAndroidApplicationId()},
--- a/source/AndroidResolver/src/PlayServicesResolver.cs
+++ b/source/AndroidResolver/src/PlayServicesResolver.cs
@@ -2155,7 +2155,8 @@ namespace GooglePlayServices {
                         // unless user has actively opted to use relative path
                         if (exportEnabled && !useRelativeRepoPath) {
                             repoUri = String.Format("\"{0}\"",
-                                                    Path.Combine(projectFileUri, repoPath));
+                                                    FileUtils.PosixPathSeparators(
+                                                      Path.Combine(projectFileUri, repoPath)));
                         } else {
                             repoUri = String.Format("(unityProjectPath + \"/{0}\")", repoPath);
                         }
--- a/source/AndroidResolver/src/SettingsDialog.cs
+++ b/source/AndroidResolver/src/SettingsDialog.cs
@@ -38,6 +38,7 @@ namespace GooglePlayServices {
             internal bool explodeAars;
             internal bool patchAndroidManifest;
             internal bool patchMainTemplateGradle;
+            internal bool useRelativeRepoPath;
             internal bool patchPropertiesTemplateGradle;
             internal string localMavenRepoDir;
             internal bool useJetifier;
@@ -59,6 +60,7 @@ namespace GooglePlayServices {
                 explodeAars = SettingsDialog.ExplodeAars;
                 patchAndroidManifest = SettingsDialog.PatchAndroidManifest;
                 patchMainTemplateGradle = SettingsDialog.PatchMainTemplateGradle;
+                useRelativeRepoPath = SettingsDialog.UseRelativeRepoPath;
                 patchPropertiesTemplateGradle = SettingsDialog.PatchPropertiesTemplateGradle;
                 localMavenRepoDir = SettingsDialog.LocalMavenRepoDir;
                 useJetifier = SettingsDialog.UseJetifier;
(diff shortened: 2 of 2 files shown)


source/AndroidResolver/src/SettingsDialog.cs
```diff
@@ -502,6 +512,27 @@ private class Settings {
...
+                GUILayout.BeginHorizontal();
+                GUILayout.Label("Use relative repository path", EditorStyles.boldLabel);
+                settings.useRelativeRepoPath =
+                    EditorGUILayout.Toggle(settings.useRelativeRepoPath);
+                GUILayout.EndHorizontal();
+
+                if (settings.useRelativeRepoPath) {
+                    GUILayout.Label(
+                        "The mainTemplate.gradle file will be patched with a relative path to " +
+                        "the Local Maven Repository. This can be used to limit file changes when " +
```
chkuang-g, line 525: Seems like this new flag only matters when GradleProjectExportEnabled is true. 

Perhaps change to something like "The mainTemplate.gradle file will be patched with a relative path to the Local Maven Repository for an exported Android project."

chkuang-g (changes requested):

chkuang-g (changes requested):
Also, worth adding a test like this

https://github.com/googlesamples/unity-jar-resolver/blob/5da4d94d75ffe0ddbf87adbcc5f0eb6e654246eb/source/AndroidResolver/test/src/AndroidResolverIntegrationTests.cs#L238

chkuang-g:
> Also, worth adding a test like this
> 
> https://github.com/googlesamples/unity-jar-resolver/blob/5da4d94d75ffe0ddbf87adbcc5f0eb6e654246eb/source/AndroidResolver/test/src/AndroidResolverIntegrationTests.cs#L238
> 
> Once added, you can run locally using
> 
> ```
> ./gradles :testAndroidResolverIntegrationTests
> ```

chkuang-g:
Sorry for the late reply and thanks for the PR!

source/AndroidResolver/src/SettingsDialog.cs
strawlink: That is correct. I've updated the description to match your suggestion.

strawlink:
Unfortunately I have not been able to run the existing integration tests locally, so I've opted to add a TODO. The gradle process simply gets stuck at 85% (`:testAndroidResolverIntegrationTestsBatchMode`).

---

As a sidenote, I have also had multiple issues with setting up the build environment, which took a while to figure out:
* Project only works with Unity 5.6.7f1.
* The build script failed on multiple steps:
  * Path to editor only works with default paths
  * Python bootstrapping (ended up commenting it out locally & just using my system install)
  I've attached the output here: [PythonBootstrapStacktrace.txt](https://github.com/googlesamples/unity-jar-resolver/files/6816037/PythonBootstrapStacktrace.txt)
  * Mono path refers to a `.bat` file, but file is an `.exe` in my clean editor installation.
  * Solutions refer to `source/unity_dlls` and `unity_dlls` - had to symlink them to inside the editor folder (5.6.7f1\Editor\Data\Managed)

chkuang-g (approved):

chkuang-g:
Thanks for the contribution!

I think this also fixed #239 as well

chkuang-g:
> Unfortunately I have not been able to run the existing integration tests locally, so I've opted to add a TODO. The gradle process simply gets stuck at 85% (`:testAndroidResolverIntegrationTestsBatchMode`).
> 
> As a sidenote, I have also had multiple issues with setting up the build environment, which took a while to figure out:
> 
> * Project only works with Unity 5.6.7f1.
> * The build script failed on multiple steps:
>   
>   * Path to editor only works with default paths
>   * Python bootstrapping (ended up commenting it out locally & just using my system install)
>     I've attached the output here: [PythonBootstrapStacktrace.txt](https://github.com/googlesamples/unity-jar-resolver/files/6816037/PythonBootstrapStacktrace.txt)
>   * Mono path refers to a `.bat` file, but file is an `.exe` in my clean editor installation.
>   * Solutions refer to `source/unity_dlls` and `unity_dlls` - had to symlink them to inside the editor folder (5.6.7f1\Editor\Data\Managed)

Ah, I think you are building this from Windows. There are a couple of issue to build EDM4U from Windows now.

strawlink:
> There are a couple of issue to build EDM4U from Windows now. You can see more information on #448

Those are indeed the same issues I've experienced 👍 

> If you cannot build EDM4U, have you ever checked if your change works for your case?

I did get it to build by making the mentioned changes, and have verified the behavior for my case locally.

source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -2140,11 +2151,12 @@ public static void SettingsDialog()
...
-                        if (exportEnabled) {
+                        // If "Export Gradle Project" setting is enabled, use absolute path
+                        // unless user has actively opted to use relative path
+                        if (exportEnabled && !useRelativeRepoPath) {
                             repoUri = String.Format("\"{0}\"",
-                                                    Path.Combine(projectFileUri, repoPath));
+                                                    FileUtils.PosixPathSeparators(
+                                                      Path.Combine(projectFileUri, repoPath)));
                         } else {
                             repoUri = String.Format("(unityProjectPath + \"/{0}\")", repoPath);
```
chkuang-g, line 2161: I did tried your change locally and I noticed that this will generate the following block in exported `build.gradle`

```
([rootProject] + (rootProject.subprojects as List)).each { project ->
        def unityProjectPath = $/file:////Users/myusername/Workspace/myprojectfolder/$.replace("\\", "/")
        maven {
            url "https://maven.google.com"
        }
        maven {
            url (unityProjectPath + "/Assets/GeneratedLocalRepo/Firebase/m2repository") // Assets/Firebase/Editor/AnalyticsDependencies.xml:18, Assets/Firebase/Editor/AppDependencies.xml:22
        }
```

Note that this is still an absolute path instead of a relative path because it combines `unityProjectPath` and the relative path. For my instance, that will be `/file:////Users/myusername/Workspace/myprojectfolder//Assets/GeneratedLocalRepo/Firebase/m2repository`

I think the right way to do is to add another String field to specify the relative path from the project root to your export folder.

For instance, given the following project structure:
- Project root: "/foo"
- Export Android project path: "/foo/bar/baz"
- A new setting `Export Project Relative Path`: "bar/baz"

That way, the code can export something like the following as a relative path:

```
([rootProject] + (rootProject.subprojects as List)).each { project ->
        def unityProjectPath = $/file:////Users/myusername/Workspace/myprojectfolder/$.replace("\\", "/")
        maven {
            url "https://maven.google.com"
        }
        maven {
            url ("../../Assets/GeneratedLocalRepo/Firebase/m2repository") // Assets/Firebase/Editor/AnalyticsDependencies.xml:18, Assets/Firebase/Editor/AppDependencies.xml:22
        }
```

chkuang-g (changes requested):

Issue #447, opened 2021-06-24
[Bug] Building an Android project in Unity 2021.2.0b1 (beta) may cause a runtime exception in the Unity Editor when looking up GetApplicationIdentifier in UnityCompat.cs.


patm1987:
### [READ] For Firebase Unity SDK issues, please report to [Firebase Unity Sample](https://github.com/firebase/quickstart-unity/issues/new/choose)

Once you've read this section and determined that your issue is appropriate for this repository, please delete this section.

### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2021.2.0b1.3029
  * External Dependency Manager version: 1.2.162
  * Source you installed EDM4U: .unitypackage
  * Features in External Dependency Manager in use: Android Resolver
  * Plugins SDK in use: Google Play
  * Platform you are using the Unity editor on: Windows

### [REQUIRED] Please describe the issue here:
After performing an Android build (and every time launching the editor after that build), Unity repeatedly logs:
```
AmbiguousMatchException: Ambiguous match found.
  at System.RuntimeType.GetMethodImplCommon (System.String name, System.Int32 genericParameterCount, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Reflection.CallingConventions callConv, System.Type[] types, System.Reflection.ParameterModifier[] modifiers) [0x00050] in <fd976842c3634855aba83404a44c7daa>:0 
  at System.RuntimeType.GetMethodImpl (System.String name, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Reflection.CallingConventions callConv, System.Type[] types, System.Reflection.ParameterModifier[] modifiers) [0x00000] in <fd976842c3634855aba83404a44c7daa>:0 
  at System.Type.GetMethod (System.String name, System.Reflection.BindingFlags bindingAttr) [0x0000e] in <fd976842c3634855aba83404a44c7daa>:0 
  at System.Type.GetMethod (System.String name) [0x00000] in <fd976842c3634855aba83404a44c7daa>:0 
  at GooglePlayServices.UnityCompat.GetUnity56AndAboveApplicationIdentifier (UnityEditor.BuildTarget buildTarget) [0x00000] in /Users/chkuang/Workspace/Git/unity-jar-resolver/source/AndroidResolver/src/UnityCompat.cs:446 
[...]

Issue #448, opened 2021-06-24
[Bug] Cannot build the android resolver on Windows with a modern version of Unity (2020.3.12f1)


patm1987:
### [READ] For Firebase Unity SDK issues, please report to [Firebase Unity Sample](https://github.com/firebase/quickstart-unity/issues/new/choose)

Once you've read this section and determined that your issue is appropriate for this repository, please delete this section.

### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2020.3.12f1
  * External Dependency Manager version: HEAD
  * Source you installed EDM4U: _____ (.unitypackage or Unity Package Manager)
  * Features in External Dependency Manager in use: _____ (Android Resolver, iOS Resolver, VersionHandler, etc.)
  * Plugins SDK in use: _____ (Firebase, Admob, Facebook, etc.)
  * Platform you are using the Unity editor on: Windows (Mac, Windows, or Linux)

### [REQUIRED] Please describe the issue here:
The base issue is:
```
PS D:\unity-jar-resolver> .\gradlew.bat build

> Configure project :
INFO: Could not find files for the given pattern(s).

FAILURE: Build failed with an exception.

* Where:
Build file 'D:\unity-jar-resolver\build.gradle' line: 55

* What went wrong:
A problem occurred evaluating root project 'playServicesResolver'.
> Unity editor executable (UNITY_EXE) not found

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
PS D:\unity-jar-resolver>
```

It's possible to get past this by changing:
```
  List<String> defaultUnityPaths =
    [(OperatingSystem.UNKNOWN): ["Unity"],
     (OperatingSystem.MAC_OSX):
     ["/Applications/Unity/Unity.app/Contents/MacOS/Unity"] +
     (new FileNameFinder()).getFileNames(
       "/", "Applications/Unity/Hub/Editor/*/Unity.app/Contents/MacOS/Unity"),
     (OperatingSystem.WINDOWS):
     ["\\Program Files\\Unity\\Editor\\Unity.exe"] +
     (new FileNameFinder()).getFileNames(
       "\\", "Program Files\\Unity\\Hub\\Editor\\*\\Editor\\Unity.exe"),
     (OperatingSystem.LINUX): ["/opt/Unity/Editor/Unity"]][operatingSystem]
```

[...]

PR #449, opened 2021-06-25
Fix for #447

patm1987:
Fixed the call to GetMethod for GetApplicationIdentifier so that the method lookup is no longer ambiguous in Unity 2021.2


joaquingrech:
May be silly question but how do I download the unity package with this fix?

patm1987:
It probably won't get generated until merged. But if you clone this PR and run:
```
./gradlew build
```

It should generate the `unitypackage` you need.

You might run into https://github.com/googlesamples/unity-jar-resolver/issues/448 if you run it on Windows. It's worth a try, but you may need to follow these steps on Linux or MacOS (I was able to build on MacOS to test this patch).

chkuang-g (approved):

Issue #452, opened 2021-07-06
[Question] How to generate :podspec parameter in Podfile using iOSResolver XML?


camphongdinh:
### [READ] For Firebase Unity SDK question, please report to [Firebase Unity Sample](https://github.com/firebase/quickstart-unity/issues/new/choose)

Once you've read this section and determined that your issue is appropriate for this repository, please delete this section.

### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2019.4
  * External Dependency Manager version: 1.2.165
  * Source you installed EDM4U: .unitypackage (.unitypackage or Unity Package Manager)
  * Features in External Dependency Manager in use: iOS Resolver (Android Resolver, iOS Resolver, VersionHandler, etc.)
  * Plugins SDK in use: Firebase, Ironsource Mediation (Firebase, Admob, Facebook, etc.)
  * Platform you are using the Unity editor on: Mac (Mac, Windows, or Linux)

### [REQUIRED] Please describe the question here:

Hi, due to the need of using old AdMob version (v7.69), I am trying to modify a released Pod "IronSourceAdMobAdapter" so it can be compatible with a higher version of "IronSourceSDK". I have downloaded and edited "IronSourceAdMobAdapter.podspec.json" as below:

```
{
  "name": "IronSourceAdMobAdapter",
  "version": "4.3.19.4",
  "summary": "AdMob Adapter",
  "description": "Use this adapter to show AdMob ads",
  "homepage": "http://www.is.com/",
  "license": {
    "type": "Commercial",
    "text": "https://platform.ironsrc.com/partners/terms-and-conditions-new-user"
  },
  "authors": {
    "IronSource": "http://www.is.com/contact/"
  },
  "source": {
    "http": "https://raw.githubusercontent.com/ironsource-mobile/iOS-adapters/master/admob-adapter/4.3.19/ISAdMobAdapter4.3.19.zip"
  },
  "source_files": "ISAdMobAdapter/ISAdMobAdapter.framework/**/*.{h,m}",
  "public_header_files": "ISAdMobAdapter/ISAdMobAdapter.framework/**/*.h",
  "preserve_paths": "ISAdMobAdapter/ISAdMobAdapter.framework",
  "platforms": {
    "ios": "9.0"
  },
  "requires_arc": true,
  "vendored_frameworks": "ISAdMobAdapter/ISAdMobAdapter.framework",
  "dependencies": {
    "IronSourceSDK": [
[...]

PR #453, opened 2021-07-14
+ Add support to "podspec" parameter for #452

camphongdinh:
Add support to "podspec" parameter. If both "podspec" and "path" is present, it will prefer "podspec" over "path" for that specific pod.

### DIFF
--- a/source/IOSResolver/src/IOSResolver.cs
+++ b/source/IOSResolver/src/IOSResolver.cs
@@ -125,11 +125,15 @@ public class IOSResolver : AssetPostprocessor {
         /// <summary>
         /// Get the path of a pod without quotes.  If the path isn't present, returns an empty
         /// string.
+        /// This also support podspec parameters. It will prefer podspec parameter over path 
+        /// if present.
         /// </summary>
         public string LocalPath {
             get {
                 string path;
-                if (!propertiesByName.TryGetValue("path", out path)) return "";
+                if (!propertiesByName.TryGetValue("podspec", out path)){
+                    if (!propertiesByName.TryGetValue("path", out path)) return "";
+                } 
                 if (path.StartsWith("'") && path.EndsWith("'")) {
                     path = path.Substring(1, path.Length - 2);
                 }
@@ -276,7 +280,8 @@ public class IOSResolver : AssetPostprocessor {
             "modular_headers",
             "source",
             "subspecs",
-            "path"
+            "path",
+            "podspec"
         };
 
         public IOSXmlDependencies() {
@@ -295,6 +300,7 @@ public class IOSResolver : AssetPostprocessor {
         ///   <iosPods>
         ///     <iosPod name="name"
         ///             path="pathToLocal"
+        ///             podspec="pathToLocalPodSpecJSON"
         ///             version="versionSpec"
         ///             bitcodeEnabled="enabled"
         ///             minTargetSdk="sdk">


camphongdinh:
@googlebot I signed it!

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -125,11 +125,15 @@ private class Pod {
...
         /// Get the path of a pod without quotes.  If the path isn't present, returns an empty
         /// string.
+        /// This also support podspec parameters. It will prefer podspec parameter over path 
+        /// if present.
         /// </summary>
         public string LocalPath {
             get {
                 string path;
-                if (!propertiesByName.TryGetValue("path", out path)) return "";
+                if (!propertiesByName.TryGetValue("podspec", out path)){
```
chkuang-g, line 134: It is better to use a different property for "podspec".

Also, do similar override in `PodFilePodLine`
https://github.com/camphongdinh/unity-jar-resolver/commit/a97b71d71ef373adb64c5688fa4f00a620189545#diff-4e150db5d98cbbe99ab461893cc478b33377658b810b34ebc72f7549670ef482R154-R157

Note that from the Cocoapod doc, ":podspec" can point to an "http" url.  So perhaps only expand the path only if it does not start with "http".

chkuang-g (changes requested):
Thank you so much for contribute to this open-source repo!

I added a comment in your PR.

Please make sure this change works for your by building it first using 
```
./gradlew buildplugin
```
And then install the local build to your project.

Shawn

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -295,6 +300,7 @@ private class IOSXmlDependencies : XmlDependencies {
         ///   <iosPods>
         ///     <iosPod name="name"
         ///             path="pathToLocal"
+        ///             podspec="pathToLocalPodSpecJSON"
```
chkuang-g, line 303: I think you can specify a url here as well
https://guides.cocoapods.org/syntax/podfile.html#pod

Perhaps use
```
/// podspec="pathToPodSpec"
```

camphongdinh:
@chkuang-g  Hi, I have updated the changes: 
 - Rename "podspec" to "podspecPath" to avoid conflicts if any
 - Support http for "podspecPath"

I have built the plugin and tested. It is working fine locally with this format in Dependencies.xml:
```
<iosPods>
    <iosPod name="IronSourceFyberAdapter" podspecPath="Assets/Plugins/ThirdParties/IronSource/Editor/LocalPod/IronSourceFyberAdapter.podspec.json">
    </iosPod>
  </iosPods>
```

Still I haven't managed to compile the source into a .tgz for use in Unity Package Manager. I will try again later.

Please let me know if you need any other changes then.

Thanks
Phong

chkuang-g:
@camphongdinh 

Thanks for the update.  I will take a look today.

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -149,7 +153,23 @@ private class Pod {
...
+                        else {
+                            outputPropertiesByName.Add("podspec", String.Format("'{0}'", Path.GetFullPath(path)));
+                        }                      
+                        // Remove "path" parameter and temporary "podspecPath" parameter
+                        outputPropertiesByName.Remove("path");  
+                        outputPropertiesByName.Remove("podspecPath");  
+                    }
+                    else {
+                        outputPropertiesByName["path"] = String.Format("'{0}'", Path.GetFullPath(path));
+                    }
```
chkuang-g, line 172: Well, it is really nice to check redundancy here but I will leave that responsibility to the person who author `Dependencies.xml`, unless you are adding additional warning here when both properties exist.

I think it is much cleaner for you and your case now to just treat them as separate properties, like this
```
                var podSpecPath = PodSpecPath;
                if (!String.IsNullOrEmpty(podSpecPath)) {
                    if(Uri.IsWellFormedUriString(podSpecPath, UriKind.Absolute)) {
                        outputPropertiesByName.Add("podspec", String.Format("'{0}'", podSpecPath));
                    }
                    else {
                        outputPropertiesByName.Add("podspec", String.Format("'{0}'", Path.GetFullPath(podSpecPath)));
                    }
                }

                var localPath = LocalPath;
                if (!String.IsNullOrEmpty(localPath)) {
                    outputPropertiesByName["path"] = String.Format("'{0}'", Path.GetFullPath(localPath));
                }

```

chkuang-g (commented):
Also, could you also help to add the documentation here about the new properties?

https://github.com/googlesamples/unity-jar-resolver/blob/master/sample/Assets/ExternalDependencyManager/Editor/SampleDependencies.xml#L55

Thank you!

PR #456, opened 2021-07-29
activating highest version in duplicate dependencies.

vimanyu:
activating only highest version in duplicate android package dependencies.

Old behavior in Gradle main template:
```
    compile 'com.android.support:support-annotations:26.1.0' // Assets/ExternalDependencyManager/Editor/TestDependencies.xml:4
    compile 'com.google.android.gms:play-services-base:12.0.1' // Assets/ExternalDependencyManager/Editor/TestAdditionalDuplicateDependencies.xml:3
    compile 'com.google.android.gms:play-services-base:15.0.1' // Assets/ExternalDependencyManager/Editor/TestAdditionalDependencies.xml:3
    compile 'com.google.firebase:firebase-app-unity:5.1.1' // Assets/Extern
```

New behavior:
```
    compile 'com.android.support:support-annotations:26.1.0' // Assets/ExternalDependencyManager/Editor/TestDependencies.xml:4
    // compile 'com.google.android.gms:play-services-base:12.0.1' // Assets/ExternalDependencyManager/Editor/TestAdditionalDuplicateDependencies.xml:3
    compile 'com.google.android.gms:play-services-base:15.0.1' // Assets/ExternalDependencyManager/Editor/TestAdditionalDependencies.xml:3
    compile 'com.google.firebase:firebase-app-unity:5.1.1' // Assets/Extern
```

Added a new integration test for this behavior and confirmed all integration tests for Android Resolver passes on Unity 2020.


source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -2191,15 +2192,45 @@ public static void SettingsDialog()
...
-                    lines.Add(String.Format(
-                        "    {0} '{1}' // {2}", includeStatement, packageSpecAndSources.Key,
-                        packageSpecAndSources.Value));
+                // Build a map of dependencies with the max version of that dependency.
+                // If different packages have different versions of the same dependency,
+                // we want to activate only the highest version but still include the
+                // other versions as commented out dependency lines.
+                var dependenciesMaxVersions = new Dictionary<string, string>();
+                foreach( var dependency in dependencies) {
+                    if(!dependenciesMaxVersions.ContainsKey(dependency.VersionlessKey))
```
chkuang-g, line 2207: nit: Use `TryGetValue()` to prevent double look-up.
nit: always use bracket {} even there is one line of code unless it can be written in one line. Similarly in the other places.

source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -2191,15 +2192,45 @@ public static void SettingsDialog()
...
+                }
+                List<Dependency> dependenciesList = dependencies.OrderBy(Dependency=>Dependency.Key).ToList();
+                foreach( var dependency in dependenciesList) {
+                    // Passing the entire list to GetPackageSpecs returns results
+                    // in an arbitrary order because it takes an IEnumerable.
+                    // Hence we cannot pass the entire list as is. We pass an element
+                    // at a time to ensure the correspondence with results.
+                    List<Dependency> singleItemList = new List<Dependency>();
+                    singleItemList.Add(dependency);
+                    var packageSpecAndSources = GetPackageSpecs(singleItemList)[0];
```
chkuang-g, line 2223: Hmm...this is now `GetPackageSpecs()` should be used.

If it is passed with multiple "Dependeny", with the same package name and version, but from different sources, this will combine the sources into a string as the dictionary value.

See this
https://github.com/googlesamples/unity-jar-resolver/blob/30b4f9292db3bd11340883dd80872c9852ae70e3/source/AndroidResolver/src/GradleResolver.cs#L114

After your change, if multiple sources (either through Dependencies.xml or through code) deps on the same package and source, you will see the following in the mainTemplate.gradle.

```
    compile 'com.google.android.gms:play-services-base:15.0.1' // Assets/Editor/FooDependencies.xml:3
    compile 'com.google.android.gms:play-services-base:15.0.1' // Assets/Editor/BarDependencies.xml:3
```

And this is going to cause issue.

You might need to refactor the API a bit to make it work.

source/AndroidResolver/test/AndroidResolverIntegrationTestsUnityProject/Assets/ExternalDependencyManager/Editor/TestAdditionalDuplicateDependencies.template
```diff
@@ -0,0 +1,5 @@
+<dependencies>
+  <androidPackages>
+    <androidPackage spec="com.google.android.gms:play-services-base:12.0.1" />
```
chkuang-g, line 3: Recommend to add something with `+` like 
```
    <androidPackage spec="com.google.firebase:firebase-app-unity:5.+">
```

chkuang-g (changes requested):

source/AndroidResolver/src/PlayServicesResolver.cs
vimanyu: Done!

source/AndroidResolver/src/PlayServicesResolver.cs
vimanyu: Great catch and I have refactored the code accordingly. Ended up creating a new function that behaves like `GetPackageSpecs` but returns all other metadata that I require about the version. I am not sure if this is the best way to do it in C# because tuples are only supported in newer versions of .NET. I used something very basic to workaround.   
Side note: It seems like we didn't hit this case of duplicate dependencies because we remove duplicates when scanning xml dependencies. By the time `GradleDependenciesLines` is called, dependencies are already de-duplicated. But your point still stands and is valid as we changed the original behvior.

source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -2191,15 +2192,45 @@ public static void SettingsDialog()
...
+                    string line = "    ";
+                    // If this is not the highest version of this dependency, add a line
+                    // but comment it out by adding leading slashes.
+                    if(dependenciesMaxVersions[dependency.VersionlessKey] != dependency.Version)
+                        line += "// ";
```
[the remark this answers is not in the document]
vimanyu: Done

source/AndroidResolver/test/AndroidResolverIntegrationTestsUnityProject/Assets/ExternalDependencyManager/Editor/TestAdditionalDuplicateDependencies.template
vimanyu: Is it just the version number that you are interested in adding to the test case or do you want to swap out play-services for firebase-app-unity too?

chkuang-g: Add a new one so that we can test for both cases.

vimanyu: Added 2 new cases to the duplicate test (a package with exact same version in a different dependencies.xml file and a package with wildcard versioning).

source/AndroidResolver/src/GradleResolver.cs
```diff
@@ -130,6 +130,54 @@ public class GradleResolver
...
+                // Convert the legacy "LATEST" version spec to a Gradle version spec.
+                var packageSpec = dependency.Version.ToUpper() == "LATEST" ?
+                    dependency.VersionlessKey + ":+" : dependency.Key;
+                var source = CommandLine.SplitLines(dependency.CreatedBy)[0];
+                string sources;
+                if (sourcesByPackageSpec.TryGetValue(packageSpec, out sources)) {
+                    sources = sources + ", " + source;
+                } else {
+                    sources = source;
+                }
```
chkuang-g, line 159: This is the same logic in GradleResolver.DependenciesToPackageSpecs().
You can make this as a separate function.

source/AndroidResolver/src/GradleResolver.cs
```diff
@@ -130,6 +130,54 @@ public class GradleResolver
...
+                versionlessKeysByPackageSpec[packageSpec] = dependency.VersionlessKey;
+            }
+            // Create a new dictionary which combines the data from sources and
+            // versions dictionaries. Key is packageSpec string and value is a
+            // three-item list of strings of sources, version less string
+            // and dependency version.
+            // NOTE: tuples are a more elegant way of doing this but they work
+            // only in .NET frameworks >= 4.0.
+            var mergedByPackageSpec = new Dictionary<string, List<string>>();
+            foreach(var item in sourcesByPackageSpec) {
```
chkuang-g, line 171: You do not need a separate loop.  You can construct mergedByPackageSpec in the previous loop.

source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -2191,15 +2203,51 @@ public static void SettingsDialog()
...
+                                 packageSpecString),
+                                 level: LogLevel.Info
+                            );
+                        line += "// ";
+                    }
+
+                    line += String.Format(
+                            "{0} '{1}' // {2}", includeStatement, packageSpecString,
+                            packageSourcesString);
+                    lines.Add(line);
```
chkuang-g, line 2250: Alright,  I think you are over-complicated logic and introduce new untested APIs here.

Here is my proposal:
`GetPackageSpecs()` returns a map from "packageSpec" to "sources"
What you need is to know which "packageSpec" should be commented.
What if you can just construct a Set of packageSpec-To-Comment. This logic should be pretty easy to integrate into your logic to find the max version..
Then you just need to test if the packageSpec is in that set.  If so, add "//". If not, don't add "//".

In this case, you do not need `GetPackageSpecsWithVersions()` nor `DependenciesToPackageSpecsWithVersions()`

chkuang-g (changes requested):

source/AndroidResolver/src/PlayServicesResolver.cs
chkuang-g: If you are making this change, ignore all my other comments

source/AndroidResolver/src/GradleResolver.cs
vimanyu: Ended up following your other suggestion and not needing this code.

source/AndroidResolver/src/GradleResolver.cs
vimanyu: Ended up following your other suggestion and not needing this code.

source/AndroidResolver/src/PlayServicesResolver.cs
vimanyu: Done! Thanks for the input :). Makes much more sense to do it this way.

source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -2191,15 +2192,56 @@ public static void SettingsDialog()
...
-                    (new Dependency.VersionComparer()).Compare("3.4", version) >= 0 ?
+                    versionComparer.Compare("3.4", version) >= 0 ?
                     "implementation" : "compile";
                 if (includeDependenciesBlock) lines.Add("dependencies {");
+                // Build a map of dependencies with the max version of that dependency.
+                // If different packages have different versions of the same dependency,
+                // we want to activate only the highest version but still include the
+                // other versions as commented out dependency lines.
+                var dependenciesMaxVersions = new Dictionary<string, Dependency>();
+                Dependency dependencyMaxVersion;
```
chkuang-g, line 2206: This variable can be in the first level of the loop below.  Otherwise, it looks like this is a variable used cross different iteration of the loop.

chkuang-g (changes requested):

source/AndroidResolver/src/PlayServicesResolver.cs
vimanyu: Done

source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -18,6 +18,7 @@ namespace GooglePlayServices {
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
```
chkuang-g, line 21: nit: remove

chkuang-g (changes requested):

source/AndroidResolver/src/PlayServicesResolver.cs
vimanyu: Good catch! Done!

chkuang-g (approved):

Issue #461, opened 2021-08-30
[FR] Add X86_64 abi support for android from Unity version 2020.3.17f1


chorobochrontochor:
### Feature proposal

* EDM4U Component: Android Resolver

Unity 2020.3.17f1 now supports X86 (again) and X86-64 abis for android platform because of ChromeOS builds but android resolver wont allow us to resolve libraries for X86-64 architecture for now

davidrogers-unity:
Here are the changes necessary to resolve this issue.

[android-x86_64.diff.zip](https://github.com/googlesamples/unity-jar-resolver/files/7092611/android-x86_64.diff.zip)

chkuang-g:
@chorobochrontochor Thank you for the reporting!

@davidrogers-unity Thank you for the suggestion! It looks like a relatively small change. Could you make a PR for this? Thank you!

Shawn

davidrogers-unity:
@chkuang-g I was going to but I'm having trouble authenticating with GitHub via CLI. Retrying via Github Desktop...

https://github.com/googlesamples/unity-jar-resolver/pull/463

chkuang-g:
Fantastic! I'll take a look in a bit

daz001:
Hello, I tried compiling the requested change and using the new compiled resolver version
I now see that in AndroidResolverDependencies.xml I correctly have:
<setting name="androidAbis" value="arm64-v8a,armeabi-v7a,x86,x86_64" />

In the build gradle file, it no longer excludes x86_64. But, if i extract the aab, i'm missing the files in:
"base\lib\x86_64"

eg. all other lib directories also contain libFirebaseCppApp-8_3_0.so & libFirebaseCppAnalytics.so & one of my own libs. but x86_64 does not contain them - is there something else i am missing?

PR #463, opened 2021-09-01
Unity now supports x86_64 for Android apps on Chrome OS. Add it to th…

davidrogers-unity:
To resolve issue #461. x86_64 support for Android projects.

### DIFF
--- a/source/AndroidResolver/src/AndroidAbis.cs
+++ b/source/AndroidResolver/src/AndroidAbis.cs
@@ -240,6 +240,7 @@ internal class AndroidAbis {
 
     /// <summary>
     /// Get / set the target device ABI (Unity >= 5.0.x)
+    /// Unity >= 2019.4 supports armeabi-v7a, arm64-v8a, x86, x86_64 & fat (i.e armeabi-v7a, arm64, x86, x86_64)
     /// Unity >= 2017.4 supports armeabi-v7a, arm64-v8a, x86 & fat (i.e armeabi-v7a, arm64, x86)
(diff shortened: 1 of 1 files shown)


davidrogers-unity:
I have signed a CLA now. But I don't know how to get it to recheck.

chkuang-g (changes requested):
Looks reasonable.  And does not seems to break anything prior to 2020.3.17 :)

Could you also update the comment here, just to keep doc up-to-date?
https://github.com/googlesamples/unity-jar-resolver/blob/41072ec47290642e126388ae3c5dd43d42695d9b/source/AndroidResolver/src/AndroidAbis.cs#L242

davidrogers-unity:
Good call and that made me think about something. We will only be supporting x86_64 from 2019.4+. Do we need to do any special handling for that? Or will everything work fine even if the architecture support doesn't exist in Unity from 2017.4-2019.3?

chkuang-g:
That probably would not be an issue since this block of code tested if the enum exists.
https://github.com/googlesamples/unity-jar-resolver/blob/master/source/AndroidResolver/src/AndroidAbis.cs#L80

Unless the enum exists in certain version Unity which has not support ChromeOS yet. :/

The worst case is that EDM4U will not [strip away unused abi](https://github.com/googlesamples/unity-jar-resolver#abi-stripping), ex. `x86_64`, which may make the the apk size bigger.

chkuang-g:
Second thought, I wonder if Android would load the correct library when both `x86` and `x86_64` libraries are presented...

davidrogers-unity:
It should, right? I think the libraries are chosen based on their directory name matching the current ABI.

chkuang-g (approved):

PR #469, opened 2021-10-06
Move versioned dlls to be under a version folder (Fix #441)

### LINKED ISSUE #441: [Known Issue] iOSResolver.dll failed to load with Unity 2021.1.11+ on Mac

  * Unity editor version: Issue occurred with `2021.1.11`, `2021.1.12` (Works with `2021.1.10` and `2021.2.0`)
  * External Dependency Manager version: Tried `1.2.160` and `1.2.165`. But version is probably irrelevant here.
  * Features in External Dependency Manager in use: iOS Resolver
  * Plugins SDK in use: None
  * Platform you are using the Unity editor on: Mac

(Please list the full steps to reproduce the issue. Include device logs, Unity logs, and stack traces if available.)

What's the issue repro rate? 100%, even with empty project.

## Issue Statement
iOS Resolver library cannot be loaded in Unity 2021.1.11 and 2021.1.12. (Did not test 2021.2.0 beta yet). It can be loaded properly with Unity 2021.1.10.

After importing EDM4U, Unity logs an error like:
```
Assembly 'Assets/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.165.dll' will not be loaded due to errors:
Assembly name 'Google.IOSResolver' does not match file name 'Google.IOSResolver_v1.2.165'
```
And iOS Resolver is not loaded, verified through the absence of `Assets > External Dependency Manager > iOS Resolver` menu item. The odd thing is that all the other libraries, ex. Android Resolver, Version Handler, are loaded with no issue.

The same issue has been reported on Unity forum [here](https://forum.unity.com/threads/google-iosresolver-will-not-be-loaded-in-2021-1-11f1-and-later.1128617/)

This seems to be a regression introduced in Unity `2021.1.11` as documented in [the release note](https://unity3d.com/unity/whats-new/2021.1.11).
- Scripting: Editor logs an Error whenever the filename of an assembly does not match its name. ([1109539](https://issuetracker.unity3d.com/issues/scriptable-object-scripts-can-not-be-found-when-the-editor-is-restarted))

## Workaround
Since the issue is the mismatched library name, a simple workaround is to rename the `dll` file.

For instance, 
- Rename `Assets/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.165.dll` to `Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll`
- Rename `Assets/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.165.dll.meta` to `Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.meta`
- Rename `Assets/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.165.dll.mdb` to `Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.mdb`
- Rename `Assets/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.165.dll.mdb.meta` to `Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.mdb.meta`

After the renaming, Unity should be able to load iOS Resolver library properly.

**However, there is a caveat**: After renaming, Version Handler will no longer able to manage those files. That is, when a newer version of EDM4U is installed, Version Handler will not be able to automatically remove the older version of iOS Resolver. 

The workaround is to remove the entire folder of `Assets/ExternalDependencyManager/` folder before upgrading.

chkuang-g:
Move the versioned dlls such as `IOSResolver.dll` to be under a version folder instead of using version postfix like `_v1.2.166`

This is a workaround for Unity 2021.1.11+ not loading `IOSResolver_v1.2.166.dll` because the filename does not match the assembly.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -290,6 +290,8 @@ project.ext {
   pipExe = new File(pythonBinDir, "pip3")
   // Python packages required by export_unity_package.py
   exportUnityPackageRequirements = ["absl-py", "PyYAML"]
+  // Python packages required by gen_guids.py
+  genGuidRequirements = ["absl-py"]
 }
 
 // Configure com.jetbrains.python.envs to bootstrap a Python install.
@@ -485,11 +487,13 @@ List<String> splitFilenameExtension(File fileObj) {
  * @param fileObj File to add version to.
  * @param fullVersionPrefix if true uses the "_version-" otherwise uses "_v-".
  * @param postfix Optional string to add before the extensioon.
+ * @param useVersionDir If true, place the file to be under a folder named after
+ *        the version number, instead of changing the filename.
  *
  * @returns File which includes an encoded version.
  */
 File versionedAssetFile(File fileObj, Boolean fullVersionPrefix,
-                        String postfix) {
+                        String postfix, Boolean useVersionDir) {
   String basename
   String extension
   (basename, extension) = splitFilenameExtension(fileObj)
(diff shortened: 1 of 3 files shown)


chkuang-g:
Note that the manifest file, ex. [this one](https://github.com/googlesamples/unity-jar-resolver/blob/master/exploded/Assets/ExternalDependencyManager/Editor/external-dependency-manager_version-1.2.166_manifest.txt), cannot be under version folder, or it will break the entire VersionHandler logic.

This PR only move dlls to the version folder.

Read [this comment](https://github.com/googlesamples/unity-jar-resolver/issues/441#issuecomment-915690686) about why those dlls were named that way before.

cynthiajoan (approved):
As discussed, once the new version is released, we need to update the following two files
https://github.com/firebase/firebase-unity-sdk/blob/main/cmake/firebase_unity_version.cmake
https://github.com/firebase/firebase-unity-sdk/blob/main/external/CMakeLists.txt

PR #491, opened 2022-01-12
Swap jcenter and mavenCentral order

legionaryu:
The jcenter is a Read-Only repository nowadays, and hence we should only use it only as a last resource.
You can learn more about the jcenter deprecation here: https://developer.android.com/studio/build/jcenter-migration


legionaryu:
I've just accepted the CLA, but I don't have any idea how to refresh the status on this PR, can someone help me?

chkuang-g:
jcenter is removed in https://github.com/googlesamples/unity-jar-resolver/pull/494.

But thank you for the PR :)

PR #500, opened 2022-01-28
Possible fix for issue (Windows) Fix #498 

### LINKED ISSUE #498: [Bug] unexpected char: '\' when using project export

  * Unity editor version: 2021.1.16f1
  * External Dependency Manager version: 1.2.168
  * Plugins SDK in use: Firebase, GoogleSignIn

1. Install Firebase from .unitypackage
2. Enable 
* Custom Main Gradle Template
* Custom Launcher Gradle Template
* Custom Gradle Properties Template
3. Export project
4. Open unityLibrary/build.gradle
5. See incorrect path in local maven url

100%

Caused by the following code in PlayServicesResolver.cs
```csharp
// If "Export Gradle Project" setting is enabled, gradle project expects
// absolute path.
if (exportEnabled) {
    repoUri = String.Format("\"{0}\"",
                            Path.Combine(projectFileUri, repoPath));
}
```
For example, `Path.Combine("file:///C:/Unity/TestGame", "Assets/GeneratedLocalRepo/Firebase/m2repository")` produces `file:///C:/Unity/TestGame\Assets/GeneratedLocalRepo/Firebase/m2repository`

paulinon:


paulinon:
After a bit of research, it appears that the behavior reported by the developer is working as intended. The [Path.Combine](https://docs.unity3d.com/ScriptReference/Path.Combine.html) method adds a [DirectorySeparatorChar](https://docs.unity3d.com/ScriptReference/Path.DirectorySeparatorChar.html) if the path does not end with a valid separator character, and a `\` is the default character for Windows.

andreimamlay:
While `\` is the default character for Windows, it is an invalid character in `File URI scheme` which is expected in gradle configuration

https://en.wikipedia.org/wiki/File_URI_scheme

PR #502, opened 2022-02-13
Speed up iOS resolver by 10-15 minutes

### LINKED ISSUE #470: [FR] Change the default cocoapods source to the CDN
### Feature proposal

* EDM4U Component: iOS Resolver

Cocoapods changed the default spec repo source to one that uses a CDN on version 1.8. If no source is defined by the Podfile, the CDN source will be the one to be used.
The EDM4U iOS cocoapods component is hardcoding the old repo: `https://github.com/CocoaPods/Specs.git`, which is significantly slower.
Updating the repo [here](https://github.com/googlesamples/unity-jar-resolver/blob/master/source/IOSResolver/src/IOSResolver.cs#L80) would be the most straightforward solution. Alternatively, we could either let the default repo be configured by users (defaulting to the CDN one), or remove that logic altogether, and let cocoapods default to the correct repo source.

Nezz:
CocoaPods has [introduced their CDN in 2019](https://blog.cocoapods.org/CocoaPods-1.8.0-beta/) and made it the default source. EDM's iOS resolver explicitly specifies the old source which has become extremely slow. The old source takes 10-15 minutes to clone on our build agents, making iOS builds needlessly slow. Using the CDN allows us to skip this step and start resolving dependencies right away.

This isn't obvious in local development because it only happens once and gets cached in `~/.cocoapods`, but disposable build agents have to run it with every single build. Caching the repo is slow because it's around a gigabyte and contains over one million files.

Resolves #470.

### DIFF
--- a/source/IOSResolver/src/IOSResolver.cs
+++ b/source/IOSResolver/src/IOSResolver.cs
@@ -77,7 +77,7 @@ public class IOSResolver : AssetPostprocessor {
         /// See: https://guides.cocoapods.org/syntax/podfile.html#source
         /// </summary>
         public List<string> sources = new List<string>() {
-            "https://github.com/CocoaPods/Specs.git"
+            "https://cdn.cocoapods.org/"
         };
 
         /// <summary>


Mazkoshak (approved):

chkuang-g (approved):

chkuang-g:
Thank you for the fix!

chkuang-g:
I'm putting together a fix for https://github.com/firebase/quickstart-unity/issues/1178. Once that is done, I'll make another release for EDM4U.

PR #508, opened 2022-04-01
Properly support pods with static library and Swift Framework

chkuang-g:
For more context [b/196153528](http://b/196153528)

A couple of changes:

1. Change `Link Framework Statically` setting default value to `true`
2. Add a workaround to properly link Swift Standard Library if any pod is using Swift Framework and `Link Framework Statically` is `true`. This workaround is turned `OFF` by default.  It basically does the following
   a. Add a `dummy.swift` file to Xcode project
   b. Set `SWIFT_VERSION` build property.  If Xcode project has multiple targets (generated by Unity 2019.3+), then also enable `CLANG_ENABLE_MODULES` and `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES`


chkuang-g:
Test with Analytics testapp, FirebaseAnalytics 8.8.0 (.unitypackage. This depends on a pod containing static library) and [Facebook Unity SDK](https://developers.facebook.com/docs/unity/) (this depends on a pod containing Swift Framework). and Xcode `13.2`. 

Due to that Facebook SDK always append `use_frameworks!` at the end of Podfile, which will be duplicated with `use_frameworks! :linkage => :static` added by EDM4U, I would need to manually move that line, run `pod install` under Xcode project folder before building Xcode app.

source/IOSResolver/src/IOSResolver.cs
```diff
@@ -1032,12 +1041,40 @@ public enum CocoapodsIntegrationMethod {
...
         get { return settings.GetBool(PREFERENCE_PODFILE_STATIC_LINK_FRAMEWORKS,
-                                    defaultValue: false); }
+                                    defaultValue: true); }
         set {
             settings.SetBool(PREFERENCE_PODFILE_STATIC_LINK_FRAMEWORKS, value);
         }
     }
 
+    /// <summary>
+    /// Ehether to enable Swift Framework support workaround.
```
AlmostMatt, line 1051: typo (Whether)

source/IOSResolver/src/IOSResolverSettingsDialog.cs
```diff
@@ -248,6 +262,33 @@ private class Settings {
...
+                GUILayout.BeginHorizontal();
+                GUILayout.Label("Enable Swift Framework Support Workaround",
+                    EditorStyles.boldLabel);
+                settings.swiftFrameworkSupportWorkaroundEnabled =
+                    EditorGUILayout.Toggle(settings.swiftFrameworkSupportWorkaroundEnabled);
+                GUILayout.EndHorizontal();
+                GUILayout.Label("This workround patches the Xcode project to properly link Swift " +
+                                "Standard Library when some plugins depend on Swift Framework " +
+                                "pods by:");
+                GUILayout.Label("1. Add a dummy Swift file to Xcode project.");
```
AlmostMatt, line 275: just to clarify - even when ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES is set the dummy swift file is still necessary?

AlmostMatt, line 275: just to clarify - even when ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES is set the dummy swift file is still necessary?

source/IOSResolver/src/IOSResolverSettingsDialog.cs
```diff
@@ -248,6 +262,33 @@ private class Settings {
...
+
+                if (settings.swiftFrameworkSupportWorkaroundEnabled) {
+                    GUILayout.BeginHorizontal();
+                    GUILayout.Label("Swift Framework Version",
+                        EditorStyles.boldLabel);
+                    settings.swiftLanguageVersion =
+                        EditorGUILayout.TextField(settings.swiftLanguageVersion);
+                    GUILayout.EndHorizontal();
+                    GUILayout.Label("Used to set 'SWIFT_VERSION' build setting in Xcode. Leave " +
+                                    "it blank to update it manually.");
```
AlmostMatt, line 288: Rather than saying "update it manually", would it be more accurate to say something like "use the "Swift Language Version" build property in the Xcode project"?

I ask because it isn't clear to me what is meant by updating it manually.

AlmostMatt, line 288: Rather than saying "update it manually", would it be more accurate to say something like "use the "Swift Language Version" build property in the Xcode project"?

I ask because it isn't clear to me what is meant by updating it manually.

source/IOSResolver/src/IOSResolverSettingsDialog.cs
chkuang-g: Yes. :(

chkuang-g: Yes :(

source/IOSResolver/src/IOSResolverSettingsDialog.cs
chkuang-g: Updated

source/IOSResolver/src/IOSResolver.cs
chkuang-g: Ah, good catch.  Thank you.
Done

a-maurice (approved):

PR #526, opened 2022-06-22
Remove the Always embed swift libraries flag

a-maurice:
As part of enabled swift, we had forced ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES to YES, but that should not be necessary, as we also add an empty swift file to the target, which will enable swift. On top of that, the always embed flag seems to cause problems for some when submitting apps because of the additional framework folder it might make.

See:
https://github.com/firebase/firebase-unity-sdk/issues/349
and
https://github.com/googleads/googleads-mobile-unity/issues/2065


chkuang-g (commented):
LGTM

chkuang-g:
Note that this solution only `avoid` stepping into this issue.  If you still encounter the same problem, please refer to 
https://forum.unity.com/threads/an-empty-folder-named-frameworks-in-unityframework-framework.751559/

PR #540, opened 2022-09-26
add tvOS structures to export_unity_package.py

DellaBitta:
Update `export_unity_package.py` with the ability to handle `tvOS` build targets.


chkuang-g (changes requested):
Need change based on previous comment

DellaBitta:
Updated. Please take a look!

chkuang-g (approved):

PR #543, opened 2022-09-29
Workaroud when `AssetDatabase.CreateFolder` failed

chkuang-g:
With certain version of Unity, ex. 2022.2.0b8 or 2023.1.0a11, `AssetDatabase.CreateFolder()` can fail when the folder name is version number like `9.0.0`. In this case, the API does not return empty guid but a guid with zeroes.

This change make sure that the folder can still be created using `Directory.CreateDirectory()`, which may not trigger AssetDatabase refresh in the older version of Unity.

### DIFF
--- a/source/VersionHandlerImpl/src/FileUtils.cs
+++ b/source/VersionHandlerImpl/src/FileUtils.cs
@@ -53,6 +53,11 @@ namespace Google {
                 new Regex(@"^(Library[/\\]PackageCache[/\\])([^/\\]+)(@[^/\\]+)[/\\](.*)?$");
 
         /// <summary>
+        /// Regex to match invalid GUID like "00000000-0000-0000-0000-000000000000" or "00000000000000000000000000000000"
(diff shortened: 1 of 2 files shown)


vangogih:
@DellaBitta hey, it's some critical bug because it reproduced in most 2022+ unity versions. Bug ticket to unity has already created but it's unknown when it will be fixed. [More info](https://github.com/firebase/quickstart-unity/issues/1272#issuecomment-1196590521)

DellaBitta (approved):

chkuang-g:
With certain version of Unity, ex. 2022.2.0b8 or 2023.1.0a11, AssetDatabase.CreateFolder() can fail when the folder name is version number like 9.0.0. In this case, the API does not return empty guid but a guid with zeroes.

This change make sure that the folder can still be created using Directory.CreateDirectory(), which may not trigger AssetDatabase refresh in the older version of Unity.

PR #549, opened 2022-10-04
Add log when Unity failed to create folder

chkuang-g:
Follow up for #543. When Unity failed to create folder using Unity API `AssetDatabase.CreateFolder()`, even with the workaround in #543, Unity would still generate two error message and there is no way to turn it off.

This change add an additional log after Unity API fails to inform the user to ignore this error.

### DIFF
--- a/source/VersionHandlerImpl/src/FileUtils.cs
+++ b/source/VersionHandlerImpl/src/FileUtils.cs
@@ -639,7 +639,7 @@ namespace Google {
         /// </summary>
         /// <param name="path">Path to the file/directory that needs checking.</param>
         /// <returns>True if all folders are created successfully.</returns>
-        public static bool CreateFolder(string path) {
+        public static bool CreateFolder(string path, Google.Logger logger = null) {
             if (AssetDatabase.IsValidFolder(path)) {
                 return true;
             }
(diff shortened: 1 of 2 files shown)


source/VersionHandlerImpl/src/FileUtils.cs
```diff
@@ -656,6 +656,20 @@ public class RemoveAssetsResult {
                 return true;
             }
 
+            if (logger != null) {
+                logger.Log(
+                    String.Format(
+                        "Please ignore Unity error message like '{0}'.\n" +
```
DellaBitta, line 662: I would just say "Please ignore Unity error messages '{0}'"

DellaBitta (approved):

source/VersionHandlerImpl/src/FileUtils.cs
chkuang-g: It is pretty counter intuitive that when Unity log an error and EDM4U says to ignore it without any explanation. 

I think this is better keep this long.

I have a feeling this is still not good enough since maybe some dev CI would expect no Error in their integration test.  But in that case, we can at least point them to the Unity bug report.

DellaBitta: It's not the length of the text, it's just a bit confusing to read.

How about `Please ignore Unity error messages similar to` ?

PR #572, opened 2022-11-04
Fix iOS Resolver to produce :tvos target podfiles

DellaBitta:
Updates to IOSResolver.cc to handle reading of `UnityEditor.PlayerSettings.tvOS.targetOSVersionString` for when active build targets are tvOS. Previously the `UnityEditor.PlayerSettings.iOS.targetOSVersionString` was used.

This work involved:

- A general breakout of `TargetSDK` into `IosTargetSDK` and `tvOSTargetSDK`.
- Renamed `TargetSDKVersion` to `TargetIosSDKVersionString` to better differentiate it from `TargetIosSDKVersionNum`, and did likewise for the new tvOS counterparts.
- Added a new poller to monitor changes to `PlayerSettings.tvOS.targetOSVersionString` akin to what we do for the PlayerSettings version value on iOS.
- In GenPodfile(), when the active build target is tvOS, populate the Podfile with `platform target :tvos` instead of `platform target :ios` and use the version string from `PlayerSettings.tvOS` instead of `PlayerSettings.iOS`.


a-maurice (commented):
I notice this is basically duplicating every function for a iOS and tvOS version.  Would it be possible to have one version, that takes in a parameter indicating if it is for iOS or tvOS?  Just thinking of the future, if we were to theoretically add something like watchOS, I wouldn't want 3 versions of every function.

a-maurice (approved):

DellaBitta:
We spoke offline. The Polling mechanism for the targetSDK values requires that TargetIosSDK and TargetTvosSDK be class properties. Condensing the TargetSDK logic into a single method would probably create even more code than reduce it, as it would simply replicate the code currently in the properties but in a new location with switch/case statements, etc.

So merging this as is for now!

PR #587, opened 2022-12-28
Avoid calling UpdateVersionedAssetsOnUpdate with each Play in Editor.

bhallionOhbibi:
Avoid unecessary computation of manifest each time the play mode is started in editor.
This computation takes multiple seconds each time we press play on many projects.
This computation is already made with each asset refresh.

Please note that this was already made on the fidebase-unity-sdk equivalent few years ago: `GenerateXmlFromGoogleServicesJson`
https://github.com/firebase/firebase-unity-sdk/blob/778a4841ac67616b2bd88a6e6f740a816d75b13b/editor/app/src/GenerateXmlFromGoogleServicesJson.cs#L96


Tommigun1980:
Please please please accept this pull request. It is completely unacceptable that the resolver makes iterating projects slower.

Issue #601, opened 2023-03-15
[FR] Optimize XmlDependencies.IsDependenciesFile for Performance


chrisyarbrough:
### Feature proposal

* EDM4U Component: Android Resolver, iOS Resolver

I've observed a significant performance slowdown in the Unity Editor when using Android Resolver and iOS Resolver in a large project with numerous assets:

<img width="937" alt="XmlDependencies_IsDependenciesFile_Performance" src="https://user-images.githubusercontent.com/17833862/225452596-5b27e872-c462-4b56-90e6-bf5ec49dee9c.png">

As seen in the profiler screenshot, the XmlDependencies.IsDependenciesFile method takes 30 seconds to execute in my current project setup. While reducing the calls to this method by refreshing the Unity AssetDatabase less frequently could be a solution, it may not always be feasible due to certain project limitations. Instead, optimizing this method seems more practical.

The performance bottleneck appears to be the regex pattern in the following code snippet:

`internal HashSet<Regex> fileRegularExpressions = new HashSet<Regex> {
    new Regex(@".*[/\\]Editor[/\\].*Dependencies\.xml$")
};

internal bool IsDependenciesFile(string filename) {
    foreach (var regex in fileRegularExpressions) {
        if (regex.Match(filename).Success) {
            return true;
        }
    }
    return false;
}`

A possible optimization involves replacing the regex pattern with simpler string manipulation methods like this example (untested):

`internal bool IsDependenciesFile(string filename) {
    const string EditorFolder = "/Editor/";
    const string DependenciesFileSuffix = "Dependencies.xml";
    int editorIndex = filename.IndexOf(EditorFolder, StringComparison.OrdinalIgnoreCase);
    
    return editorIndex >= 0 && filename.EndsWith(DependenciesFileSuffix, StringComparison.OrdinalIgnoreCase);
}`

However, PlayServicesResolver also uses fileRegularExpressions to add additional patterns. I'm unsure how this can be easily accommodated, but I propose the following potential improvements:

- Replace PlayServicesResolver patterns with simpler string methods, if possible
- Change fileRegularExpressions from a collection of Regex objects to a delegate or filter object that can be either a Regex or a simpler string manipulation
- Examine the regex pattern and options for possible enhancements

I would be happy to investigate this issue myself and test changes in my project. Any feedback is greatly appreciated!

chkuang-g:
I see your point. 

[...]

PR #602, opened 2023-03-17
Improve performance of XmlDependencies IsDependenciesFile

chrisyarbrough:
### Why this change?

This implements the improvement for issue: https://github.com/googlesamples/unity-jar-resolver/issues/601

The `XmlDependencies.IsDependenciesFile` method is called thousands of times in large projects and is now much faster than before:

IsDependenciesFile_Before

IsDependenciesFile _After

The previous slowdown was causing long waiting times during content iteration in Unity projects with more than a few hundred assets.

### What did I do?

I added a new test project to add a test which fixates the previous logic of file pattern matches with regex. I then replaced the regex with simpler string manipulation.

To avoid breaking API changes, I kept the public method in PlayServicesResolver the same, but internally, the new code uses a delegate which accepts a file path and returns true if it's a match. This way we can have both the Regex expressions passed from existing user code and a faster version for the internal default which only searches for the Dependencies.xml files in editor folders.

### Review & Release

The review should be easiest commit-by-commit.

Feedback is greatly appreciated and please let me know which steps I need to take to make my change release-ready or if any of the maintainers will take over. Thank you! :)

**Tasks**
- [x] Implement and unit-test my change
- [x] Test the change in a production project
- [x] Update the changelog in the repo root
- [x] Update the version in the build.gradle
- [ ] Wait for PR feedback
- [ ] Build the release using `./gradlew release`
- [ ] Create release commit using `./gradlew gitCreateReleaseCommit`
- [ ] Tag the release using `./gradlew gitTagRelease`and `git push --tag REMOTE HEAD:master`

### DIFF
--- a/source/AndroidResolver/src/XmlDependencies.cs
+++ b/source/AndroidResolver/src/XmlDependencies.cs
@@ -41,7 +41,7 @@ namespace GooglePlayServices {
             if (!filename.EndsWith("Dependencies.xml")) {
                 return false;
             }
-
+            // This method was optimized to avoid using regex after profiling results from a large Unity project.
             return filename.Contains("/Editor/") || filename.Contains(@"\Editor\");
         }
 


source/AndroidResolver/src/XmlDependencies.cs
```diff
@@ -46,13 +38,9 @@ internal class XmlDependencies {
...
-        internal bool IsDependenciesFile(string filename) {
-            foreach (var regex in fileRegularExpressions) {
-                if (regex.Match(filename).Success) {
-                    return true;
-                }
-            }
-            return false;
+        internal static bool IsDependenciesFile(string filename) {
+            bool isInEditorFolder = filename.Contains("/Editor/") || filename.Contains(@"\Editor\");
+            return isInEditorFolder && filename.EndsWith("Dependencies.xml");
```
chkuang-g, line 43: This is string parsing for 3 times (worst cast)
Also, I expect there are far more files to be under `Editor` folder than a file name ended with `Dependencies.xml`
Here is my recommendation:
1. Early out when the filename does not end with `Dependencies.xml`
2. Otherwise, use regex to check if the file is under `Editor` folder. sometime like `".*[/\\]Editor[/\\].*"`

Regex is suppose to match at `O(n)` time so that should be better than parsing strings twice just to know if it is under `Editor` folder.

source/AndroidResolver/src/FileMatchPattern.cs
```diff
@@ -0,0 +1,3 @@
+namespace GooglePlayServices {
+	public delegate bool FileMatchPattern(string filename);
```
chkuang-g, line 2: I am not entirely sure the benefit to introduce this delegate and the change in `PlayServicesResolver.cs`. I do not think there would be any performance gain, is there?

source/AndroidResolverTests/AndroidResolverTests.csproj
chkuang-g, line 1: Kudos to you to add a Unit test for this! 

Could you move the entire test to `source/AndroidResolver/unit_tests` for consistency sake. Thank you!

Also, please add the following lines to `build.gradle`, like right...[here](https://github.com/googlesamples/unity-jar-resolver/blob/5f2051d122f8dedfe52a334fdcdab2007019ebec/build.gradle#L1772).

```
// TODO: Add NUnit tests for #602
```

Currently there are some issue with this build script to run NUnit tests from Unity 2019. While we do not want to burden you with the effort to fix it, adding a line here as a reminder would be nice.

source/AndroidResolverTests/XmlDependenciesTests.cs
chkuang-g, line 1: Could you move this to `source/AndroidResolver/unit_tests/src` folder for consistency sake.

Tommigun1980:
Hi @chkuang-g, when do you estimate this fix to land? Thanks.

Tommigun1980:
@chkuang-g Hi, when will this PR land?

source/AndroidResolver/src/XmlDependencies.cs
chrisyarbrough: Thanks, the early-out definitely makes it even faster. 

However, as for the regex, that's the point of the PR: while profiling my large company project I noticed that the regex was taking considerable time compared to the more simple string comparisons. In theory, they should be the same, but the practical overhead appears to be quite substantial, even if pre-compiling the regex instruction.

source/AndroidResolver/src/FileMatchPattern.cs
chrisyarbrough: I agree, removed it!

source/AndroidResolverTests/AndroidResolverTests.csproj
```diff
@@ -0,0 +1,63 @@
...
+		<ErrorReport>prompt</ErrorReport>
+		<WarningLevel>4</WarningLevel>
+	</PropertyGroup>
+	<ItemGroup>
+		<Reference Include="System" />
+		<Reference Include="System.Core" />
+		<Reference Include="System.Data" />
+		<Reference Include="System.Xml" />
+		<Reference Include="nunit.framework, Version=3.5.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb">
+			<HintPath>..\packages\NUnit.3.5.0\lib\net45\nunit.framework.dll</HintPath>
```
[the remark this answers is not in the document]
chrisyarbrough: I think I copied this from a previous version of JarResolverTests, basically just added a nuget package reference to NUnit to run the tests in the Rider IDE outside of Unity. I'm not sure what the "packages" directory here is, but removing the HintPath doesn't prevent me from running the tests.

Please advise if there's a way to run the test that's better aligned with the rest of the project. My local project on macOS in Rider doesn't really compile all too well, and I'm used to much newer SDK versions of dotnet, so I'm not sure what the best way would be in this project. :)

chrisyarbrough:
I finally got some time to apply the review feedback. I retested the changes in my large company project and the custom changes still show some good improvements in Unity.

Issue #594, opened 2023-01-23
[Bug] Android Gradle project repositories block don't work for the latest Unity versions


256p:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2022.2.2f1
  * External Dependency Manager version: 1.2.175
  * Source you installed EDM4U: .unitypackage
  * Features in External Dependency Manager in use: Android Resolver
  * Plugins SDK in use: None
  * Platform you are using the Unity editor on: Mac

### [REQUIRED] Please describe the issue here:
(Please list the full steps to reproduce the issue. Include device logs, Unity logs, and stack traces if available.)

#### Steps to reproduce:
1. Add Dependencies.xml which has custom `<repository>` specified.
2. Enable custom main gradle template and custom grade properties template.
3. Run Android resolver through  Assets > External Dependency Manager > Android Resolver > Resolve or Force Resolve.
#### Expected:
Successful build
#### Observed:
A lot of warnings from Gradle like that:
```
Build was configured to prefer settings repositories over project repositories but repository 'maven' was added by build file 'unityLibrary/build.gradle'
```
And this error:
```
Execution failed for task ':launcher:checkReleaseDuplicateClasses'.
> Could not resolve all dependencies for configuration ':launcher:releaseRuntimeClasspath'.
  The project declares repositories, effectively ignoring the repositories you have declared in the settings.
  You can figure out how project repositories are declared by configuring your build to fail on project repositories.
  See https://docs.gradle.org/7.2/userguide/declaring_repositories.html#sub:fail_build_on_project_repositories for details.
   > Could not find com.ironsource.sdk:mediationsdk:7.2.7.
     Searched in the following locations:
       - https://dl.google.com/dl/android/maven2/com/ironsource/sdk/mediationsdk/7.2.7/mediationsdk-7.2.7.pom
       - https://repo.maven.apache.org/maven2/com/ironsource/sdk/mediationsdk/7.2.7/mediationsdk-7.2.7.pom
       - file:/Users/stanislavspetkevics/Desktop/unity-jar-resolver-custom-template/Library/Bee/Android/Prj/Mono2x/Gradle/unityLibrary/libs/mediationsdk-7.2.7.jar
       - file:/Users/stanislavspetkevics/Desktop/unity-jar-resolver-custom-template/Library/Bee/Android/Prj/Mono2x/Gradle/unityLibrary/libs/mediationsdk.jar
     Required by:
         project :launcher > project :unityLibrary
[...]

PR #605, opened 2023-03-29
Support for Gradle dependencyResolutionManagement RepositoriesMode.PREFER_SETTINGS

256p:
This PR fixes https://github.com/googlesamples/unity-jar-resolver/issues/594#issue-1553387077

Changes influence only builds when main Gradle template patching is enabled.

Changes:
* Added "Patch settingsTemplate.gradle" checkbox in Android Resolver settings.
* Implemented repositories injection in `settingsTemplate.gradle`
* Added `/resolve/gradlesettingstemplate` report URL to the analytics. (Not sure if it is needed)

The logic of repositories injection depends on the project state:
* Project state when `mainTemplate.gradle` and `settingsTemplate.gradle` are enabled and mainTemplate patching and settingsTemplate patching in Android Resolver settings are enabled:
  1. Android Resolver will check if `settingsTemplate.gradle` contains `dependencyResolutionManagement` block with `RepositoriesMode.PREFER_SETTINGS` or `RepositoriesMode.FAIL_ON_PROJECT_REPOS`. If the match is successful it would mean that Gradle is configured to read repositories from `settingsTemplate.gradle` and will not read repositories specified in `mainTemplate.gradle`.
  2. If the previous check is successful then Android Resolver will inject repositories in `settingsTemplate.gradle` before `dependencyResolutionManagement` line. Injected repositories are in separate `dependencyResolutionManagement` block and look like this:  
  ```
  // Android Resolver Repos Start
  dependencyResolutionManagement {
      repositories {
          def unityProjectPath = $/file:///**DIR_UNITYPROJECT**/$.replace("\\", "/")
          maven {
              url "https://maven.google.com"
          }
          maven {
              url "https://maven.google.com/" // Assets/GoogleMobileAds/Editor/GoogleMobileAdsDependencies.xml:7, Assets/GoogleMobileAds/Editor/GoogleMobileAdsDependencies.xml:12, Assets/GoogleMobileAds/Editor/GoogleMobileAdsDependencies.xml:17, Assets/GoogleMobileAds/Editor/GoogleUmpDependencies.xml:7
          }
          mavenLocal()
          mavenCentral()
      }
  }
  // Android Resolver Repos End
  ```
  3. If the previous check isn't successful then Android Resolver will inject repositories the same way as it is doing in older versions in `mainTemplate.gradle`
* In the project state when `mainTemplate.gradle` and `settingsTemplate.gradle` are enabled and mainTemplate patching is **enabled** but settingsTemplate patching in Android Resolver settings is **disabled**, Android Resolver will not patch `settingsTemplate.gradle` this will lead to the build fail and the user will need to manage repositories manually. 
* Project state when `mainTemplate.gradle` is enabled and mainTemplate patching and settingsTemplate patching in Android Resolver settings are enabled but `settingsTemplate.gradle` is disabled:
  1. If the Unity version is equal to or above 2022.2.10 Android Resolver will fail and in the error log will be asked to enable the settings template in "Project Settings". 
  2. If the Unity version is below 2022.2.10 then Android Resolver will check if the default `settingsTemplate.gradle` in `PlaybackEngines/AndroidPlayer/Tools/GradleTemplates` contains `dependencyResolutionManagement` block. Effectively this will match any Unity version equal to or above the 2022.2 version because `dependencyResolutionManagement` was introduced in the 2022.2 version.
    2.1. If the previous check didn't match then Android Resolver will work as in the previous versions.
    2.2. If the previous check is successful then Android Resolver will enable `settingsTemplate.gradle` automatically (by copying the default template). Decided to do it in such a way because Unity  >=2022.2 and <2022.2.10 uses a settings template for repositories but don't have an option in "Project Settings" to enable a custom settings template. This option appeared only in 2022.2.10.


chkuang-g:
Hi @256p
Thank you so much for contributing this patch for EDM4U!
We are doing very similar things in #610 with some additional touches

1. Unity 2022.2.1 - 2022.2.9 does not support template variable like `**DIR_UNITYPROJECT**` unfortunately. Therefore we added two settings to force EDM4U to always use absolute path.
2. Unity 2022.2.1 - 2022.2.9 also does not have an option to enable custom Gradle settings template. This change automatically copy the template from the engine folder.

Still we appreciate your effort to put together this change for the community. And kudos to you to also add integration tests! The proper fix for this is a bit involved so we spent some time to make sure the experience is right.

I will close this PR for now.

Regards,
Shawn

PR #609, opened 2023-04-10
Inject Maven Repo to settingTemplate.gradle from 2022.2+

chkuang-g:
Add a new feature and settings for Unity 2022.2+ in Android Resolver to enable patching `Assets/Plugins/Android/settingsTemplate.gradle` for Maven repository specified in `Dependencies.xml`.


source/AndroidResolver/src/PlayServicesResolver.cs
```diff
@@ -1975,13 +1995,23 @@ public class AndroidSdkRootChangedArgs : EventArgs {
...
                 lastError = "";
                 RunOnMainThread.Run(() => {
-                        finishResolution(GradleTemplateResolver.InjectDependencies(
-                            PlayServicesSupport.GetAllDependencies().Values) &&
-                            patchGradleProperties(), lastError);
+                        var dependencies = PlayServicesSupport.GetAllDependencies().Values;
+                        finishResolution(
+                            GradleTemplateResolver.InjectDependencies(dependencies) &&
+                            patchGradleProperties() &&
+                            patchSettingsTemplateGradle(dependencies), lastError);
```
a-maurice, line 2014: I'm a bit confused by what lastError is meant to do.  Like, if the patchSettingsTemplateGradle changes lastError, I don't think it will affect the variable that is passed into finishResolution (I could be wrong, passing parameters around like this is definitely confusing).

source/AndroidResolver/src/GradleTemplateResolver.cs
```diff
@@ -527,5 +588,180 @@ private class TextFileLineInjector {
                              "Gradle Template", "gradletemplate",
                              injectors, resolutionMeasurementParameters);
         }
+
+        /// <summary>
```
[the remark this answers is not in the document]
chkuang-g: Ah, good catch. Added.

source/AndroidResolver/src/PlayServicesResolver.cs
chkuang-g: I agree this is hard to understand. 

In fact, this actually works (Just tried it). I think this is probably because the function parameters are evaluated in order. Therefore, `patchGradleProperties()` and `patchSettingsTemplateGradle(dependencies)` has already been called before `lastError` is passed to `finishResolution()`.

a-maurice (approved):

source/AndroidResolver/src/PlayServicesResolver.cs
a-maurice: Hmm, I guess that does make sense.  Definitely a bit confusing, but I don't think it is worth blocking for.

PR #613, opened 2023-04-15
`Explode Aar` option applies to all cases

### LINKED ISSUE #287: disabling Explore AAR still modifies AARs after downloading
disabling Explode AAR still modifies AARs after downloading in 1.2.129 (current).

to reproduce:
1. uncheck Explode AAR in options, enable verbose
2. start resolve
3. output log contains stuff about replacing applicationId (and it is actually replaced in the files)

### LINKED ISSUE #584: [Bug] disabling Explore AAR still downloads and modifies AARs

  * Unity editor version: 2020.3
  * External Dependency Manager version: 1.2.175
  * Plugins SDK in use: Admob

Disable "Explode AARs" in Android Resolver Settings and then make Force Resolve. 
AARs are still downloaded and exploaded to the Plugins/Android folder.

chkuang-g:
Before this change, `Explode Aar` option in Android Resolver settings only applies when `Export Project` build setting is enabled, while Android Resolver always explode Aar libraries when the build setting is disabled.

This changes `Explode Aar` option to be applied no matter `Export Project` build setting is enabled or not. It is still recommended to enable this option for older version of Unity (5 to 2018) since certain version of Unity may generate broken Gradle project.

Fixes #584
Fixes #287

### DIFF
--- a/source/AndroidResolver/src/GradleResolver.cs
+++ b/source/AndroidResolver/src/GradleResolver.cs
@@ -1161,7 +1161,7 @@ namespace GooglePlayServices
             // To work around this when Gradle builds are enabled, explosion is enabled for all
             // AARs that require variable expansion unless this behavior is explicitly disabled
             // in the settings dialog.
-            if (PlayServicesResolver.GradleProjectExportEnabled && !SettingsDialog.ExplodeAars) {
+            if (!SettingsDialog.ExplodeAars) {
                 return false;
             }
             // If this version of Unity doesn't support AAR files, always explode.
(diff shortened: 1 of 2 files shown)


chkuang-g:
@doterax
@ievgenmukhin
@wei-jiaxing 
@shefich

Please comment if you have any concern about this change.

doterax (approved):
Everything looks fine for me

source/AndroidResolver/src/GradleResolver.cs
```diff
@@ -1161,7 +1161,7 @@ internal static bool SupportsAarFiles
             // To work around this when Gradle builds are enabled, explosion is enabled for all
```
a-maurice, line 1161: Update the comments for this?

a-maurice (approved):

source/AndroidResolver/src/GradleResolver.cs
chkuang-g: This comment is actually good with current code. It does not explicitly mention about "Gradle Project Export need to be enabled as well" part.

a-maurice: Ah, sounds good

PR #633, opened 2023-08-15
Add flag for "use_modular_headers!" Podfile setting

yanmasharski:


TheoSabattie:
@a-maurice / @stewartmiles / @chkuang-g / (...) Can you review/validate this MR please?

Applovin requires to disable static-framework.  
By disabling this option, firebase does not work anymore until this flag is added to the podfile.  
I'll be benefit to add this option to resolve the issue directly from unity-jar-resolver instead of making workaround.

Thanks in advance.

Théo

PR #648, opened 2023-10-23
Update Test and Build workflow

chkuang-g:
- Prevent unintended labels from triggering test workflow. Only `tests-requested` label would trigger test workflow now.
- Prevent pull request from triggering build workflows since test
  workflow would build EDM4U.
- Prevent test workflow from emitting error message just because
  `grep` command did not get any failed tests.
- Add info for failed NUnit tests

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -66,7 +66,31 @@ jobs:
             echo "include_test_modules=" >> $GITHUB_OUTPUT
             echo "exclude_test_modules=" >> $GITHUB_OUTPUT
             echo "exclude_tests=" >> $GITHUB_OUTPUT
+
+            if [[ "${{ github.event_name }}" == "schedule" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -66,8 +66,35 @@ jobs:
             echo "include_test_modules=" >> $GITHUB_OUTPUT
             echo "exclude_test_modules=" >> $GITHUB_OUTPUT
             echo "exclude_tests=" >> $GITHUB_OUTPUT
+
+            if [[ "${{ github.event_name }}" == "schedule" ]]; then
```
a-maurice, line 70: I think a comment explaining that this is currently checking for invalid triggers would be good, cause it looks a bit odd by itself.

chkuang-g: Done

a-maurice (approved):

Issue #524, opened 2022-06-12
[Bug] GvhProjectSettings.xml changes/resets itself


ManuelRauber:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2021.3.4f1
  * External Dependency Manager version: 1.2.171
  * Source you installed EDM4U: .unitypackage
  * Features in External Dependency Manager in use: Android Resolver
  * Platform you are using the Unity editor on: Mac

### [REQUIRED] Please describe the issue here:
Sometimes, not really reproducible, the file GvhProjectSettings.xml changes/resets itself. This is super annoying, because the file pops up in the git stage:

![image](https://user-images.githubusercontent.com/740791/173223351-222b8a45-76f9-4791-b167-2786213f97bd.png)

As well as this dialogs pops up all the time

![image](https://user-images.githubusercontent.com/740791/173223366-de6cff03-636a-4f13-8328-9ab987a77d24.png)

And as you can see from the diff file above, the settings get reset, e.g. I've disabled both auto-resolution on build and in editor and it will reset back to true, because that's the default.

I currently can not reproduce it exactly, when it happens. Sometimes it does not happen all day, sometimes it happens multiple times day.

paulinon:
Hi @ManuelRauber,

Could you provide the steps you've performed before facing this issue so that we can identify what's causing this behavior?

faveris:
I have the same issue. I've just installed External Dependency Manager with one of SDKs and these things start to happen. Looks like it happens after recompilation but not after each one.

Unity editor version: 2021.3.3f1
External Dependency Manager version: 1.2.169
Source you installed EDM4U: .unitypackage
Features in External Dependency Manager in use: Android Resolver
Platform you are using the Unity editor on: Mac

CriticalForce:
We are facing same issue on
Unity: 2021.3.4f1
Editor: Mac (Silicon)

ManuelRauber:
> Hi @ManuelRauber,
> 
> Could you provide the steps you've performed before facing this issue so that we can identify what's causing this behavior?

Unfortunately, I can not answer this, because I simply use the Unity Editor. 
Sometimes I just open it and then I see a change in git and the file has been written as shown in the first post.

Sometimes that does not happen, I have no clear way to reproduce this behavior. 

Currently, I'm carefully looking whenever this change happens, but I've not found a clear way yet.

belike81:
Having the exact same issue. Reset happens every few minutes.
[...]

PR #652, opened 2023-11-06
Fix project settings file getting corrupted

AliAlbarrak:
# Why this change?
This resolve the settings file corruption as described in #524

Even though reading and writing of the settings file is protected by a lock, it is possible (due to _domain reload_) that 2 separate running processes are trying to access the file at the same time so the lock in one process is not aware of the other process.

A scenario like this could happen.
1. Code in domain A will start writing to ProjectsSettings file
2. Code in domain B will start reading the file
3. If domain A didn't finish writing yet, domain B will encounter unexpected EOF
4. Domain B will assume the file is broken and reset settings
5. Domain B will write default settings to ProjectsSettings file

# What did I do?
I made the writing happen to a temporary file. When the writing is complete, copy the temp file to the target path. This avoids the file being in an "invalid state" for a short time during the writing.

# Review
Feedback is greatly appreciated and please let me know which steps I need to take to make my change release-ready.

### DIFF
--- a/source/VersionHandlerImpl/src/ProjectSettings.cs
+++ b/source/VersionHandlerImpl/src/ProjectSettings.cs
@@ -257,8 +257,9 @@ namespace Google {
         {
             string stringValue = value.ToString();
 
-            if (!settings.ContainsKey(name) || settings[name] != stringValue) 
+            if (!settings.ContainsKey(name) || settings[name] != stringValue) {
                 IsModified = true;
+            }
             
             settings[name] = stringValue;
         }
--- a/source/VersionHandlerImpl/src/ProjectSettings.cs
+++ b/source/VersionHandlerImpl/src/ProjectSettings.cs
@@ -817,9 +817,10 @@ namespace Google {
                                       PROJECT_SETTINGS_FILE), LogLevel.Error);
                     return;
                 }
+                string tmpFile = Path.GetTempFileName();
                 try {
                     using (var writer =
-                           XmlWriter.Create(PROJECT_SETTINGS_FILE,
+                           XmlWriter.Create(tmpFile,
                                             new XmlWriterSettings {
                                                 Encoding = new UTF8Encoding(false),
                                                 Indent = true,
@@ -839,6 +840,8 @@ namespace Google {
                         }
                         writer.WriteEndElement();
                     }
+                    
+                    File.Copy(tmpFile, PROJECT_SETTINGS_FILE, true);
                 } catch (Exception exception) {
                     if (exception is IOException || exception is UnauthorizedAccessException) {
                         logger.Log(String.Format("Unable to write to '{0}' ({1}, " +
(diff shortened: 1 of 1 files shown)


AliAlbarrak:
As reported [here](https://github.com/googlesamples/unity-jar-resolver/issues/524#issuecomment-1808595188), this fix  fails sometimes so I'll draft the pull request until I find a better solution

AliAlbarrak:
The settings file gets written multiple times with every domain reload/compile, even when settings didn't change. I added an extra check to skip file write if settings didn't change since last read.

AliAlbarrak:
Hello @a-maurice @chkuang-g and team
I'd appreciate it if you can guide me through what I need to do to get this reviewed and merged. This PR is associated with the top voted 👍 issue #524. I believe it will be great UX improvement for most users to get it fixed.

cmcpasserby:
really can non one review this and get it merged in? This issue has been going on for years and makes working with this very annoying and tedious

Tommigun1980:
I tested this pull request and my findings are at https://github.com/googlesamples/unity-jar-resolver/issues/524#issuecomment-1895834683

source/VersionHandlerImpl/src/ProjectSettings.cs
```diff
@@ -234,8 +253,14 @@ internal class InMemorySettings : ISettings {
         /// </summary>
         /// <param name="name">Name of the value.</param>
         /// <param name="value">Value to set.</param>
-        private void Set<T>(string name, T value) {
-            settings[name] = value.ToString();
+        private void Set<T>(string name, T value)
+        {
+            string stringValue = value.ToString();
+
+            if (!settings.ContainsKey(name) || settings[name] != stringValue)
```
TyelorD, line 260: To match the current coding style,readability, and portability, this if statement should have enclosing curly braces around the set call for IsModified.

TyelorD (changes requested):
I think this pull looks good to me, aside from the one change I requested.

That said, I'm not affiliated with Google in any way.

source/VersionHandlerImpl/src/ProjectSettings.cs
```diff
@@ -234,8 +253,14 @@ internal class InMemorySettings : ISettings {
...
         /// <param name="name">Name of the value.</param>
         /// <param name="value">Value to set.</param>
-        private void Set<T>(string name, T value) {
-            settings[name] = value.ToString();
+        private void Set<T>(string name, T value)
+        {
+            string stringValue = value.ToString();
+
+            if (!settings.ContainsKey(name) || settings[name] != stringValue) 
+                IsModified = true;
```
TyelorD, line 261: To match the current coding style,readability, and portability, this if statement should have enclosing curly braces around the set call for IsModified.

berniegp:
Hi! I still got settings corruption with your version. However I think I found the source of the problem and a fix here #678.

alaa13212:
I guess this is no longer needed since #678 was merged 🎉

AliAlbarrak:
I guess this is no longer needed since https://github.com/googlesamples/unity-jar-resolver/pull/678 was merged 🎉

PR #678, opened 2024-03-09
Fix project settings resetting on domain reload

### LINKED ISSUE #524: [Bug] GvhProjectSettings.xml changes/resets itself

berniegp:
This resolves #524 by avoiding to needlessly save `GvhProjectSettings.xml` on every domain reload.

In all 3 modified files, the line
```
VerboseLoggingEnabled = VerboseLoggingEnabled;
```
does 3 things on domain reload because of `[InitializeOnLoad]`:
1. Loads `GvhProjectSettings.xml` if not already done
2. Writes `GvhProjectSettings.xml` unconditionally
3. Sets `logger.Level`

This PR removes step 2) which is unneeded and the cause of #524.

It's potentially possible that writing `GvhProjectSettings.xml` as a result of user interaction might still in some cases have a race condition. I didn't investigate this. However it's much less problematic than intermittent problems on domain reload.

# Why the problem happens
The current code is correctly protecting critical sections, including saving `GvhProjectSettings.xml`, with `lock (classLock)`. However a problem occurs with [parallel asset importing enabled in Unity](https://docs.unity3d.com/Manual/ParallelImport.html). In this case there are other Unity Editor processes aside from the "main" one that take care of parallel imports. These other processes each have their own separate instance of `classLock`.

This results in multiple processes reading and writing to `GvhProjectSettings.xml` on domain reload. Eventually one of them reads the file while it's being written and ends up with empty settings from `Google.ProjectSettings.Load` which it then re-saves. As the screenshot in #524 shows, when corruption happens it's always one or more `VerboseLoggingEnabled` that remain in the settings.

## About asset import worker processes
The command line of asset import worker processes looks like:
```
"C:\tools\Unity-Hub\2022.3.19f1\Editor\Unity.exe" "-adb2" "-batchMode" "-noUpm" "-name" "AssetImportWorker7" "-projectPath" "C:/dev/unity-project" "-logFile" "Logs/AssetImportWorker7.log" "-srvPort" "64205"
```
The `-adb2` argument is unique to asset import worker processes so could be used to identify them as needed.

# Build on Windows
Binaries here https://github.com/berniegp/unity-jar-resolver/releases/tag/v1.2.179

I could not build the plugin as it stands. The (reverted) commit ff03ffda690943f17800d3a26b97ba0d6e753c33 has the changes I needed to do to `build.gradle`.

As far as I can see, the build process [prefers Unity 5.6](https://github.com/googlesamples/unity-jar-resolver/blob/master/build.gradle#L163). This version (obviously) isn't supported on Apple Silicon so I tried on Windows.

Steps for successful build:
1. Install [Unity 5.6](https://unity.com/releases/editor/archive) with Android and iOS support
2. Start Unity 5.6 to sort out activation
3. Apply ff03ffda690943f17800d3a26b97ba0d6e753c33
4. Download and extract https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_windows-x64_bin.zip to c:\tools
5. As administrator (so Python installs correctly, but pollutes add/remove programs) build with:
```
$env:JAVA_HOME='C:\tools\openjdk-11.0.2_windows-x64_bin\jdk-11.0.2\'; .\gradlew build -PUNITY_EXE=’<PATH_TO>\Unity-Hub\5.6.7f1\Editor\Unity.exe’; Remove-Item Env:\JAVA_HOME
```

Not sure if I'm doing something wrong regarding the build?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -126,7 +126,7 @@ project.ext {
     "MONO_EXE", false, {
       unityRootDirTree.matching {
         include (operatingSystem == OperatingSystem.WINDOWS ?
-          "**/bin/mono.exe" : "**/bin/mono")
+          "**/bin/mono.bat" : "**/bin/mono")
         exclude unitySearchDirExcludes
       }
     })
@@ -318,9 +318,9 @@ project.ext {
   pythonBootstrapDir = new File(externalToolsDir, "python_bootstrap")
   pythonBinDir = new File(new File(pythonBootstrapDir, "python"), "bin")
   // Python binary after it has been bootstrapped.
-  pythonExe = new File(new File(pythonBootstrapDir, "python"), "python")
+  pythonExe = new File(pythonBinDir, "python3")
   // Pip binary after it has been bootstrapped.
-  pipExe = new File(new File(new File(pythonBootstrapDir, "python"), "Scripts"), "pip3")
+  pipExe = new File(pythonBinDir, "pip3")
   // Python packages required by export_unity_package.py
   exportUnityPackageRequirements = ["absl-py", "PyYAML", "packaging"]
   // Python packages required by gen_guids.py
--- a/build.gradle
+++ b/build.gradle
@@ -126,7 +126,7 @@ project.ext {
     "MONO_EXE", false, {
       unityRootDirTree.matching {
         include (operatingSystem == OperatingSystem.WINDOWS ?
-          "**/bin/mono.bat" : "**/bin/mono")
+          "**/bin/mono.exe" : "**/bin/mono")
         exclude unitySearchDirExcludes
       }
     })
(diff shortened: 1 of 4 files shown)


berniegp:
Update: 1 month in at my org we're not seeing setting file corruption anymore on Windows and MacOS.

harlan:
What are the steps to try to get @berniegp 's https://github.com/googlesamples/unity-jar-resolver/pull/678 tested/merged etc? is that that you need to sign this CLA thing or something?

berniegp:
I already signed the CLA thing, so next steps aren't on me :)

image

argzdev:
Hey thanks for the PR, let me get some of our engineering folks to take a look into this.

a-maurice (approved):

a-maurice:
Thanks for the fix, and the detailed write up actually explaining what the issue was, that is definitely a weird case so really appreciate that you tracked it down.  I will work on getting a release out with this fix.

As for building it yourself, it can definitely be a bit finicky and depends on a bunch of tools.  I wasn't aware of a Unity 5.6 bias, the tests we have run with Unity 2019, but I do know that Apple Silicon can have issues because it depends on a specific python version which isn't available.

Issue #428, opened 2021-03-24
[Bug] High GC Allocation in Editor


snorrsi:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2020.2.6
  * External Dependency Manager version: 1.2.164
  * Source you installed EDM4U: Unity Package Manager
  * Features in External Dependency Manager in use: Android iOS, iOS Resolver
  * Plugins SDK in use: Firebase
  * Platform you are using the Unity editor on: Mac and Windows

### [REQUIRED] Please describe the issue here:
During Editor profiling it's clear that same functions in EDM4U are run multiple time which are causing high GC Alloc due to string ToLower function.
Due to string conversion the GC Alloc is 10kb or so each frame or at least very often.
Can't this data be cached as the UnityEngine.Environment.CommandLine surely does not change often?

[source/VersionHandlerImpl/src/ExecutionEnvironment.cs#L32](https://github.com/googlesamples/unity-jar-resolver/blob/3dfc84aba31425c1def8a2da7027757d4b3d59fd/source/VersionHandlerImpl/src/ExecutionEnvironment.cs#L32
)source/VersionHandlerImpl/src/ExecutionEnvironment.cs
31 | public static bool InBatchMode {
32 | get { return Environment.CommandLine.ToLower().Contains("-batchmode"); }
… |  
39 | get { return Environment.CommandLine.ToLower().Contains("-executemethod"); }
40 | }

[source/VersionHandlerImpl/src/Logger.cs#L52](https://github.com/googlesamples/unity-jar-resolver/blob/62d83320541e4bcb28cab93b57c99838120f42a2/source/VersionHandlerImpl/src/Logger.cs#L52
)source/VersionHandlerImpl/src/Logger.cs
50 | internal static bool DebugLoggingEnabled {
51 | get {
52 | return Environment.CommandLine.ToLower().Contains("-gvh_log_debug");
52 | }

#### Please answer the following, if applicable:
What's the issue repro rate? 100%

What happened? How can we make the problem occur?
During Editor profiling it's clear that same functions in EDM4U are run multiple time which are causing high GC Alloc due to string ToLower function.
Run a project in Unity Editor with Call Stacks (GC Alloc) profiling and look at origination of GC Allocation in Editor.

Attached is callstack and image from the Unity Profiling. All of the 5.8kb GC Alloc is related to this issue.

[...]

PR #680, opened 2024-04-19
Update ExecutionEnvironment to cache command line

snorrsi:
Simple fix for unnecessary GC Allocation using command line

See #428

### DIFF
--- a/source/VersionHandlerImpl/src/ExecutionEnvironment.cs
+++ b/source/VersionHandlerImpl/src/ExecutionEnvironment.cs
@@ -26,17 +26,34 @@ namespace Google {
 internal class ExecutionEnvironment {
 
     /// <summary>
+    /// Cached lower case CommandLine
+    /// </summary>
+    private static string Cached_Environment_CommandLine_Lower = null;
+    
+    /// <summary>
+    /// Get/Set cached lower case CommandLine
+    /// </summary>
+    public static string Environment_CommandLine_Lower {
+        get {
+            if (Cached_Environment_CommandLine_Lower != null) {
+                Cached_Environment_CommandLine_Lower = Environment.CommandLine.ToLower();
+            }
+            return Cached_Environment_CommandLine_Lower;
+        }
+    }
+    
+    /// <summary>
     /// Whether the editor was started in batch mode.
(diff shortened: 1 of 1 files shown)


arcv (approved):

source/VersionHandlerImpl/src/ExecutionEnvironment.cs
```diff
@@ -25,26 +25,43 @@ namespace Google {
 /// </summary>
 internal class ExecutionEnvironment {
 
+    /// <summary>
+    /// Cached lower case CommandLine
+    /// </summary>
+    private static string Cached_Environment_CommandLine_Lower = null;
```
Tommigun1980, line 31: Not a constant so name should not user underscore notation.

source/VersionHandlerImpl/src/ExecutionEnvironment.cs
```diff
@@ -25,26 +25,43 @@ namespace Google {
...
+            return Cached_Environment_CommandLine_Lower;
+        }
+    }
+    
     /// <summary>
     /// Whether the editor was started in batch mode.
     /// </summary>
     public static bool InBatchMode {
-        get { return Environment.CommandLine.ToLower().Contains("-batchmode"); }
+        get { return Environment_CommandLine_Lower.Contains("-batchmode"); }
```
Tommigun1980, line 49: This is better than before but couldn't you just read all three of the switches it's interested in inside a constructor and just return booleans in the properties? This still runs unnecessary string operations every frame.

Tommigun1980 (changes requested):

Tommigun1980:
This has to be merged in. Is the jar resolver maintained anymore or has Google sunset Unity support? The editor is constantly allocating memory which makes long-running automated test running inside editor slow down over time etc. Of course you can't read the command line every frame, read it once like this pull request does. Merge it in now please.

Tommigun1980 (approved):
Approving nevertheless just to try to get Google to merge this in.

Tommigun1980:
@paulinon @a-maurice Could you just approve this as it is not ok that the resolver constantly allocates memory every frame.

Issue #622, opened 2023-05-31
[Bug] IOSResolver.dll tries to load on Windows/Linux - causes error


CJxD:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2021.3.16f1
  * External Dependency Manager version: 1.2.176 (longstanding issue with all previous versions)
  * Source you installed EDM4U: .unitypackage (.unitypackage or Unity Package Manager)
  * Features in External Dependency Manager in use: Android Resolver (Android Resolver, iOS Resolver, VersionHandler, etc.)
  * Plugins SDK in use: none (Firebase, Admob, Facebook, etc.)
  * Platform you are using the Unity editor on: Linux (Mac, Windows, or Linux)

### [REQUIRED] Please describe the issue here:

1. Import the .unitypackage into any project and run on Linux (or Windows)
2. Note the following error in the console:

```
Assembly 'Assets/ExternalDependencyManager/Editor/1.2.176/Google.IOSResolver.dll' will not be loaded due to errors:
Unable to resolve reference 'UnityEditor.iOS.Extensions.Xcode'. Is the assembly missing or incompatible with the current platform?
```

#### Please answer the following, if applicable:
What's the issue repro rate? (eg 100%, 1/5 etc)

100%

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

This is because this DLL can only be loaded on a Mac. I've fixed this manually for some time by changing the Platform Settings for that file from `Any OS` to `macOS`, which updates the meta and prevents issues.

I've put up with this for a while, but now one of my customers asked how to fix the error after importing External Dependency Manager. I think it should be like this by default to prevent confusion.

chkuang-g:
I think what @CJxD is a valid fix and should be quick enough to resolve by modifying the [export configuration](https://github.com/googlesamples/unity-jar-resolver/blob/master/export_unity_package_config.json#L18) to something like the following

```
        {
        "importer": "PluginImporter",
        "platforms": [],
        "labels": ["gvhp_targets-editor"],
        "paths": [
            "ExternalDependencyManager/Editor/*/Google.IOSResolver.*"
        ],
        "override_metadata": {
          "PluginImporter": {
            "platformData": [ {
                "first" : {
                  "Editor": "Editor"
                },
                "second": {
                  "settings": {
                    "OS": "OSX"
                  }
                }
              }
            ]
          }
        },
        "override_metadata_upm": {
            "PluginImporter": {
[...]

Issue #412, opened 2020-11-09
[BUG] Unable to resolve reference 'UnityEditor.iOS.Extensions.Xcode' on Windows


DominatorDuck:
### [REQUIRED] Please fill in the following fields:

  * Unity editor version: 2020.1.11f1
  * External Dependency Manager version: 1.2.161
  * Source you installed EDM4U: .unitypackage
  * Features in External Dependency Manager in use: Android, ResolverVersionHandler
  * Plugins SDK in use: None
  * Platform you are using the Unity editor on: Windows

### [REQUIRED] Please describe the issue here:
Given a vanilla 3D project in Unity 2020.1.11f1, upon pressing play I get the following error log:

```
Assembly 'Packages/com.google.external-dependency-manager/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.161.dll' will not be loaded due to errors:
Unable to resolve reference 'UnityEditor.iOS.Extensions.Xcode'. Is the assembly missing or incompatible with the current platform?
Reference validation can be disabled in the Plugin Inspector.
```

Steps to reproduce:
- Open a Unity project (in Unity 2020.1.11f1) with the External Dependency Manager package.
- Press the play button.
- Notice the error log.

Notes:
- To create this project I originally made a project for Unity version 2019.4.8f1 and the upgraded it to 2020.1.11f1.
- The error does no occur with the very same project in Unity 2019.4.8f1.
- The only non-default package in the project is the External Dependency Manager.
- The Unity 2020.1.11f1 installation has only Windows (IL2CPP) and Android Build support.
- External Dependency Manager settings are as they come out of the package installation.
- Enabling Verbose Logging does not seem to give any more relevant information.
- Changing Build platform (from Windows to Android) seems to have no effect on the bug.

#### Please answer the following, if applicable:
What's the issue repro rate? 100%

What happened? How can we make the problem occur?

The Assembly 'Packages/com.google.external-dependency-manager/ExternalDependencyManager/Editor/Google.IOSResolver_v1.2.161.dll' fails to load.

You can reproduce the issue by following the repro steps with the proper Unity installation by using this sample project:
[https://drive.google.com/file/d/1llavcZxSVk7icuCpLv65QPrnjgV81u93/view?usp=sharing](url)

arcv:
For quick fix, you can install "iOS Build Support" module from Unity Hub. That fixes the issue.

DominatorDuck:
[...]

PR #691, opened 2024-06-17
Fix IOSResolver platforms. Use only on mac

RevenantX:
Set correct platforms for IOSResolver (only mac editor)
Fixes: 

- https://github.com/googlesamples/unity-jar-resolver/issues/622 
- https://github.com/googlesamples/unity-jar-resolver/issues/412

Since you cannot fully build IOS builds on windows this change is totally safe

### DIFF
--- a/plugin/Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.meta
+++ b/plugin/Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.meta
@@ -10,9 +10,30 @@ PluginImporter:
   serializedVersion: 2
   iconMap: {}
   executionOrder: {}
+  defineConstraints: []
   isPreloaded: 0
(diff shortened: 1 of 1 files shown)


StephenHodgson:
@RevenantX actually you can build mac/iOS on windows/linux, just not IL2CPP for mac standalone, but you can def export iOS projects. The real fix can be found in #714

RevenantX:
@StephenHodgson you can build Mac on Windows (Mono) but not iOS. Because iOS is 64bit only - you need IL2CPP and XCode to actually build iOS application.
image
As you see Build button is grayed.

StephenHodgson:
> As you see Build button is grayed.

Build and Run is greyed out yes, but you can still build and export the xcode project itself.
Regardless your comment about being able to build on windows is incorrect. I do this all the time.
image

Issue #699, opened 2024-07-25
[Bug][Android] Build failure after 2021.3.41f1 (AGP updated from 4.2.2 to 7.4.2)


manugildev:
### [READ] For Firebase Unity SDK issues, please report to [Firebase Unity Sample](https://github.com/firebase/quickstart-unity/issues/new/choose)

Once you've read this section and determined that your issue is appropriate for this repository, please delete this section.

### [REQUIRED] Please fill in the following fields:

  * Unity editor version: `2021.3.41f1`
  * External Dependency Manager version: `1.2.178`
  * Source you installed EDM4U: `.unitypackage`
  * Features in External Dependency Manager in use: `Android Resolver`
  * Plugins SDK in use: `Firebase 12.1.0`
  * Platform you are using the Unity editor on: `Mac`

### [REQUIRED] Please describe the issue here:
- Download [Unity 2021.3.41f1](unityhub://2021.3.41f1)
- Create an empty project and switch target platform to Android ([failing_edm4u_2021_3_41f1.zip](https://drive.google.com/file/d/1BSYJIrem-FXOKSKjfauHqbSrgJ1N0lRH/view?usp=drive_link))
- Import latest [google-mobile-ads-sdk](https://github.com/googleads/googleads-mobile-unity/releases/tag/v9.2.0) (9.2.0)
- Import latest [firebase-unity-sdk](https://github.com/firebase/firebase-unity-sdk/releases/tag/v12.1.0) (12.1.0)
- `Assets > External Dependency Manager > Android Resolver > Resolve`
- Build

#### Please answer the following, if applicable:
What's the issue repro rate? 100%

What happened? How can we make the problem occur?
`.apk` doesn't build because of gradle failures.
```
FAILURE: Build completed with 3 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':launcher:checkReleaseDuplicateClasses'.
> Could not resolve all files for configuration ':launcher:releaseRuntimeClasspath'.
   > Could not find com.google.firebase:firebase-analytics-unity:12.1.0.
     Searched in the following locations:
       - https://dl.google.com/dl/android/maven2/com/google/firebase/firebase-analytics-unity/12.1.0/firebase-analytics-unity-12.1.0.pom
       - https://repo.maven.apache.org/maven2/com/google/firebase/firebase-analytics-unity/12.1.0/firebase-analytics-unity-12.1.0.pom
[...]

PR #700, opened 2024-08-01
use AGP version as condition check

argzdev:
Possible fix for #699 

Adding the plugin repositories in `settingsTemplate.gradle` and `mainTemplate.gradle` condition checker will be based on the AGP version instead of the Unity Editor version.

### DIFF
--- a/source/AndroidResolver/src/GradleTemplateResolver.cs
+++ b/source/AndroidResolver/src/GradleTemplateResolver.cs
@@ -14,7 +14,8 @@
 //    limitations under the License.
 // </copyright>
 
-namespace GooglePlayServices {
+namespace GooglePlayServices
+{
     using System;
     using System.Collections.Generic;
     using System.IO;
@@ -28,7 +29,8 @@ namespace GooglePlayServices {
     /// <summary>
     /// Resolver which simply injects dependencies into a gradle template file.
     /// </summary>
-    internal class GradleTemplateResolver {
+    internal class GradleTemplateResolver
+    {
 
         /// <summary>
         /// Filename of the Custom Gradle Properties Template file.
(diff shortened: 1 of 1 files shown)


manugildev:
In order to test, please download `2021.3.42f1` from [`unityhub://2021.3.42f1/f1197811e8ce`](unityhub://2021.3.42f1/f1197811e8ce) (copy paste into your browser navigation bar).

argzdev:
Thanks for the heads up, @manugildev. From my understanding, shouldn't this fix also work for version `2021.3.41f1` since the mapped AGP version is >= 7.0?

manugildev:
> Update: We're running some tests with this fix, but it still encounters the same issue.

Yes, `2021.3.41f1` had a bug where **DIR_UNITYPROJECT** on the `settingsTemplate.gradle` is not being replaced by the actual directory. This might be causing an issue. It's fixed for `2021.3.42f1`, but as you said, since it's officially released next week that version is still missing the JDK requirements.

While we wait for `2021.3.42f1` to be fully publicly released next week, could you check in `2021.3.41f1` if manually replacing `**DIR_UNITYPROJECT**` in `settingsTemplate.gradle` by the actual directory of the project solves the issue?

manugildev:
> Update: We're running some tests with this fix, but it still encounters the same issue.

@argzdev  Yes, `2021.3.41f1` had a bug where `**DIR_UNITYPROJECT**` on the `settingsTemplate.gradle` is not being replaced by the actual directory. This might be causing an issue. It's fixed for `2021.3.42f1`, but as you said, since it's officially released next week that version is still missing the JDK requirements.

While we wait for `2021.3.42f1` to be fully publicly released next week, could you check in `2021.3.41f1` if manually replacing `**DIR_UNITYPROJECT**` in `settingsTemplate.gradle` by the actual directory of the project solves the issue?

Just to clarify there are two issues with `2021.3.41f1` (AGP 7.4.2) at the moment:
 - The repositories section needs to be put in `settingsTemplate.gradle`, this is solved by your changes in unity-jar-resolver.
 - `**DIR_UNITYPROJECT**` is not being replaced by the actual directory during build/export time. This is solved by Unity in next release 2021.3.42f1.  

Thank you for your help!

a-maurice:
I ended up doing this change with this PR, so that I could get the release out faster with it.  https://github.com/googlesamples/unity-jar-resolver/pull/702

Thanks for setting it up, and testing it out on your end.

a-maurice:
@manugildev Good to hear.  I can't share any firm dates, but the next releases should have the latest EDM4U version.  The Firebase Unity SDK should be out soon.

PR #743, opened 2025-05-19
Turn off validateReferences for the iOS Resolver

a-maurice:
By default, DLLs imported into Unity will validate all their references at load time. The iOS Resolver depends on the Unity iOS Support Library being present, which is not true by default for Windows developers. By turning off validateReferences, the library will no longer throw an exception at load time, but instead at runtime, which the library itself can handle instead.

https://github.com/googlesamples/unity-jar-resolver/issues/412
https://github.com/googlesamples/unity-jar-resolver/issues/622

### DIFF
--- a/plugin/Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.meta
+++ b/plugin/Assets/ExternalDependencyManager/Editor/Google.IOSResolver.dll.meta
@@ -12,6 +12,7 @@ PluginImporter:
   executionOrder: {}
   isPreloaded: 0
   isOverridable: 0
+  validateReferences: 0
   platformData:
   - first:
       Any: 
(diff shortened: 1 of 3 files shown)


cynthiajoan (approved):

StephenHodgson:
This did not fix the problem. 

Tried with Unity 6000.0.49f1 with only Android and Windows IL2CPP modules installed. 

```
Assembly 'Packages/com.google.firebase.app/Firebase/Editor/Firebase.Editor.dll' will not be loaded due to errors:
Unable to resolve reference 'UnityEditor.iOS.Extensions.Xcode'. Is the assembly missing or incompatible with the current platform?
Reference validation can be disabled in the Plugin Inspector.
```

a-maurice:
@StephenHodgson Odd.  In your project, is "Validate References" still enabled for the Google.IOSResolver.dll?  This change should make it off by default, but that error message seems like it is still on for you.

image

PR #756, opened 2025-10-22
Add support for using Swift Packages

a-maurice:
Adds support in the iOS Resolver to depend on Swift Packages, instead of Cocoapods.


source/IOSResolver/src/IOSResolver.cs
```diff
@@ -2148,14 +2192,26 @@ internal static void AddDummySwiftFile(
         File.WriteAllText(pbxprojPath, project.WriteToString());
     }
 
+    [PostProcessBuildAttribute(35)]
```
AustinBenoit: Nit: Can we replace 35 with a variable like "BUILD_ORDER_RES_SPM" and add a definition on line 416
```suggestion
    [PostProcessBuildAttribute(BUILD_ORDER_RES_SPM)]
```

AustinBenoit (approved):

